Issue #36, opened 2014-06-19
Vim does not get detected when run via `make` / use `tmux set-env`?


blueyed:
I have a Makefile task that starts vim, but then `:TmuxPaneCurrentCommand` is `make`.

Maybe there's some way to look into the current tmux command's subprocess, which would be `vim` then, or another method to notify the tmux pane that it should forward the keys?!

Maybe `tmux set-environment` could be called by the vim part to indicate which pane(s) are meant to receive the keys from tmux?!

PR #37, opened 2014-06-19
Use the tmux environment for communication

blueyed:
This is a proof of concept, and would need more documentation
adjustments if it works out.

This makes it less specific to Vim and avoids the fragile part of
detecting if Vim is running in the current pane.

TODO:

 - [ ] replace/escape % from tmux' pane_id in variable name

Fixes: https://github.com/christoomey/vim-tmux-navigator/issues/36

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 Vim Tmux Navigator
 ==================
 
-This plugin is a repackaging of [Mislav Marohnić's][] tmux-navigator
-configuration described in [this gist][]. When combined with a set of tmux
-key bindings, the plugin will allow you to navigate seamlessly between
-vim and tmux splits using a consistent set of hotkeys.
+This plugin is based on [Mislav Marohnić's][] tmux-navigator configuration
+described in [this gist][]. When combined with a set of tmux key bindings, the
+plugin will allow you to navigate seamlessly between Vim and tmux splits using
+a consistent set of hotkeys.
 
 **NOTE**: This requires tmux v1.8 or higher.
 
@@ -63,9 +63,6 @@ bind -n C-k if-shell "$is_vim" "send-keys C-k" "select-pane -U"
 bind -n C-l if-shell "$is_vim" "send-keys C-l" "select-pane -R"
 bind -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
 ```
-
-Thanks to Christopher Sexton who provided the updated tmux configuration in
-[this blog post][].
 
 Configuration
 -------------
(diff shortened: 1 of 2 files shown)


sethwoodworth:
It is worth noting that the grep -iqE method doesn't work with Neovim.  But this PR does work with neovim.

christoomey:
@sethwoodworth I'm surprised to hear that the `grep -iqE` method wouldn't work with Neovim? The specific pattern, `grep -iqE "(^|\/)g?(view|n?vim?)(diff)?$"` includes an optional leading "n" and is tested in the `pattern-check` file in the root of the repo. Can you comment?

Regardless, I still need to wrap my head around this alternative method and decide it its the right thing moving forward.

blueyed:
I've updated / rebased this PR/branch - adding a fixup-commit to prefer looking at the environment.

With tmux 2.0 there is `if -F`, which allows to expand FORMATS without invoking a shell.  If this could be used it should make it more efficient.

blueyed:
Does it make sense to keep the `grep` approach, given my previous comment about `tmux select-pane` failing with `sudo`/`ssh` anyway (https://github.com/christoomey/vim-tmux-navigator/pull/37#issuecomment-51611619)?

/cc @aschrab

aschrab:
No matter what's decided here I think I'll be sticking with the checking pane titles as I mentioned in a [previous comment](https://github.com/christoomey/vim-tmux-navigator/pull/37#issuecomment-51538782) rather than using the tmux environment.

Although @blueyed is correct that this doesn't really solve the problem with sudo or ssh, it does have a different failure mode. My current method still allows me to use the usual bindings to switch between vim windows on a remote system, but fails to switch between tmux panes. Since I use vim windows more than tmux panes, this works better for me than if the breakage was reversed which I believe would be the case with this change.

blueyed:
JFI: `%` is not valid in variable names for Zsh, and should therefore get replaced probably.
I've only noticed this now when adding `eval $(tmux switch-client \; show-environment -s)` to my shell config.

blueyed:
Rebased it on master.

keith:
Any news on this one? #bump :smile:

christoomey:
No news. I've yet to really test this unfortunately. @keith have you run with this locally? Does it solve the issue of the plugin not being able to tell if Vim is active (for instance due to spawning Vim from a git commend, meaning the `pane_current_command` will be `Git`)? I've also been experimenting with https://github.com/christoomey/vim-tmux-navigator/issues/96 as an alternative approach that seems dicey due to platform inconsistencies, but intriguing for its possible completeness. Thoughts?

blueyed:
Closing this in favor of #117.

keith:
I'll test out 117 instead!

PR #64, opened 2015-02-21
Add link back to Github project in tmux conf in README

ssadler:
Just a suggestion :-)


christoomey:
:+1: Do you mind squashing down to a single commit with a message like "Add link back to project in readme"?

christoomey:
@ssadler thanks for the addition. I went ahead and squashed / rebased onto master and merged this in.

PR #65, opened 2015-02-28
when current tmux pane is zoomed in, avoid accidentally zoom out by moving off the edge of vim pane.

zhengwy888:
First pull request for me on github. Let me know if you have any comment.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -40,11 +40,18 @@ function! s:NeedsVitalityRedraw()
   return exists('g:loaded_vitality') && v:version < 704 && !has("patch481")
 endfunction
 
+function! s:TmuxIsZoomed() 
+   return system("tmux display-message -p '#{window_flags}' |grep -c Z") != 0
+endfunction
+
 function! s:TmuxAwareNavigate(direction)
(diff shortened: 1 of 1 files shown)


christoomey:
Hello @zhengwy888, thanks for update and congrats on your first PR! Unfortunately I personally rely on the fact that navigating unzooms. Similarly, I rarely zoom the vim pane, but instead zoom a test runner or shell pane, and then I only do it briefly to read some scrollback and then immediately unzoom. I think spending a lot of time in a zoomed pane is a pattern to avoid, rather than enable.

All that said, as a potential alternate solution you can use "zooming" within vim itself if you need to see more of a buffer. I use the following mappings to allow me to maximize the current buffer (`<leader>-`), and then rebalance to make all buffers equal (`<leader>=`).

``` vim
" automatically rebalance windows on vim resize
autocmd VimResized * :wincmd =

" zoom a vim pane, <C-w>= to re-balance
nnoremap <leader>- :wincmd _<cr>:wincmd \|<cr>
nnoremap <leader>= :wincmd =<cr>
```

Hope that helps and thanks for sharing.

blueyed:
See #133 for a fresh approach at this.

PR #68, opened 2015-03-16
Switches to Plugin from Bundle references https://github.com/gmarik/Vund...

w1zeman1p:
Plugin is the newer version of Bundle.

references: https://github.com/gmarik/Vundle.vim/commit/0521de95eac09298c4e71b3662839612280c1ae9

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -38,7 +38,13 @@ install the plugin:
 Add the following line to your `~/.vimrc` file
 
 ``` vim
-Bundle 'christoomey/vim-tmux-navigator'
+Plugin 'christoomey/vim-tmux-navigator'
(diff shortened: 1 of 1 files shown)


Keithbsmiley:
This might not be the right place for this but as there have become so many plugin managers that occasionally update their interface, and users already normally having and knowing how to use the manager of their choice, I've taken to use saying `If you don't have a preferred installation method check out [foo]`, so things like this aren't necessary.

christoomey:
@w1zeman1p thanks for the update. @Keithbsmiley I agree in general, but am currently in between Vundle and vim-plug so I don't have a firm answer on that.

PR #74, opened 2015-05-25
fire FocusLost autocommand after switching tmux pane

0arthur:
For things like save on lost focus etc.


christoomey:
Hey @0arthur, sorry for letting this sit so long. Can you provide any additional detail as to why this is needed? I would expect something like `FocusLost` to be managed by Vim and I am wary of adding explicit handling of this in the plugin, but would consider it if I better understood the context. Thanks!

0arthur:
Vim doesn't fire the FocusLost event in a terminal. I've since seen that there are other plugins to manage this i.e. vim-tmux-focus-events and I take your point that it's not really your plugin's job, so I guess this can be dropped.

christoomey:
@0arthur thanks for the quick reply, and thanks for pointing out vim-tmux-focus-events. I was not aware of.

PR #75, opened 2015-05-26
Add readme notes for slow shell troubleshooting

christoomey:


montlebalm:
Here's my take on the new language. There are minor changes to punctuation, formatting, and wording. I ballpark'd the 80 chars, so that'll need to be rebalanced.

```
If you find that navigation within Vim (from split to split) is fine, but Vim 
to another non-Vim tmux pane is delayed, it might be due to a slow shell startup. 
Consider moving code from your shell's non-interactive rc file (e.g., `~/.zshenv`) 
into the interactive startup file (e.g., `~/.zshrc`) as Vim only sources the 
non-interactive config.
```

Issue #27, opened 2014-04-01
Tmux/Tmate/Vim navigation not working


pzula:
Hello!

Thank you for this great package!

I love this plugin, but have recently noticed an issue when using [tmate](http://tmate.io) that my keybindings don't seem to work.

When I run `:TmuxPaneCurrentCommand`, I get this output:
`protocol version mismatch (client 8, server 7)`

This is pane-ful (get it?) when using tmate to pair with others remotely, as I've grown accustomed to navigating in vim normally while in tmux. Do you have any suggestions on how I can work around this? I am using Mac OSX Mavericks, homebrew, and iTerm.

jondkinney:
I'm running Yosemite and tried to downgrade to tmux 1.8. When you start tmux it says something to the effect of: "warning this version is unsupported, please upgrade to 1.9a". I also use tmuxinator which uses some of the new `-c` stuff (I guess?) and so it didn't work either. Hopefully we can get tmate upgraded to 1.9a soon, but unfortunately it looks like the author hasn't had time for the last year to update it https://github.com/nviennot/tmate/issues/33

PR #81, opened 2015-08-07
Make everything work with tmate

notahat:
This provides a way of setting some config to better support tmate, even when the tmux and tmate versions don't match.

This should address #27 in a more helpful way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,14 +188,32 @@ Consider moving code from your shell's non-interactive rc file (e.g.,
 `~/.zshenv`) into the interactive startup file (e.g., `~/.zshrc`) as Vim only
 sources the non-interactive config.
 
-### It Doesn't Work in tmate
+### It Doesn't Work in [tmate][]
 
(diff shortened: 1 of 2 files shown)


christoomey:
@notahat Thanks a ton for this and my sincere apologies for letting it sit this long without responding!

I'm a bit sad to see that the only fix needed is to simply use the `tmate` executable, but happy to know that we now have a solution. Thanks for posting this!

While the overall solution works, I've created another PR https://github.com/christoomey/vim-tmux-navigator/pull/89 which detects the executable (`tmux` vs `tmate`). It seems to work in my local testing, but I was hoping you could pull it down and test on your machine to confirm.

christoomey:
Hey @notahat, thanks again for writing up the original summary on this. I've gone ahead and merged https://github.com/christoomey/vim-tmux-navigator/pull/89 which provides the same behavior, but automatically determines the executable. I'm going to close this as I think PR #89 covers it, but please reopen or comment if you think I missed anything.

Thanks again!

Issue #87, opened 2015-10-14
Support custom tmux socket


thejohnfreeman:
I'm on a shared machine where /tmp is cleaned regularly. I keep my tmux socket in my home directory so that I don't lose the session:

```
$ tmux -S ~/.tmux.socket
```

Every tmux command needs that argument to work:

```
$ tmux select-pane -L
failed to connect to server
$ tmux -S ~/.tmux.socket select-pane -L
# selects pane
```

vim-tmux-navigator does not work in this tmux session until I add the socket parameter in the vim script [here](https://github.com/christoomey/vim-tmux-navigator/blob/master/plugin/tmux_navigator.vim#L56). I'm not too familiar with VimScript. Ideally there'd be a plugin option (or auto-detection if possible) to control the presence and value of the socket parameter. I believe you can add that feature better than I can, or else I'd make the pull request myself. Please?

christoomey:
Hey @thejohnfreeman, this makes sense, but unfortunately it is far enough of an edge case (and not one I experience) that I am unlikely to work on implementing support.

That said, I'd be happy to work with you if you wanted to undertake the dev / documentation effort. I expect this would require:

- Extract a function for all `tmux` commands (currently 2 in the plugin file, search for `system` to find them) to centralize them
- Introduce a `g:vtn_socket` valriable
- Make the `tmux` function from the first step conditionally include the `g:vtn_socket` value if present, otherwise compose a normal tmux command
- Document the `g:vtn_socket` variable.

Please let me know if you have any other questions.

PR #88, opened 2015-10-17
Read socket path from $TMUX

thejohnfreeman:
Fixes: #87

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -19,8 +19,18 @@ function! s:InTmuxSession()
   return $TMUX != ''
 endfunction
 
+function! s:TmuxSocket()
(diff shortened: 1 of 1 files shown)


christoomey:
@thejohnfreeman this looks great! Far simpler than I'd imagined and the automatic detection is nice. I've pulled this down locally and will run with it for a few days just to confirm there are no edge cases, but this looks solid and I expect I'll merge in a few days.

Thanks!

thejohnfreeman:
Great! It was simpler than I expected too. I didn't know about `$TMUX` until I started researching for this patch.

christoomey:
Hey @thejohnfreeman, I've run without issue with your branch for the past week, so I just merged it.

Thanks again for pushing this through!

PR #89, opened 2015-10-25
Detect tmux vs tmate

christoomey:
This solves the issues related to using vim-tmux-navigator in tmux & tmate by detecting the current executable (`tmux` vs `tmate`) and explicitly running against it.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -11,6 +11,19 @@ if !exists("g:tmux_navigator_save_on_switch")
   let g:tmux_navigator_save_on_switch = 0
 endif
 
+function! s:TmuxOrTmateExecutable()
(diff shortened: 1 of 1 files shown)


karlbright:
:+1: Would love to see this merged.

christoomey:
@karlbright would you mind running with it for a few days to help me verify it's working as expected? I've been using for a bit now, but would love to get at least one other set of eyes on it.

Assuming you're using vundle, vim-plug, or similar, you can clone this and checkout the branch, then symlink it into your `~/.vim/bundle` directory.

karlbright:
@christoomey Yup, i've been running this for the last week now and have not run in to any issues. I've used tmate 4-5 times within the week and it seems to be working well.

christoomey:
@karlbright I'll take it! I've merged this into master, so you should be able to go back to a normal checkout. Thanks for commenting and testing!

PR #94, opened 2015-11-05
Not trying to save non-files to disk

JackDanger:
If the current buffer isn't backed by persistence (i.e. it doesn't have
a filename) then don't run the `update` function.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -60,7 +60,8 @@ function! s:TmuxAwareNavigate(direction)
   " a) we're toggling between the last tmux pane;
   " b) we tried switching windows in vim but it didn't have effect.
   if tmux_last_pane || nr == winnr()
-    if g:tmux_navigator_save_on_switch
+    " Unly save the file if it has a filename
+    if g:tmux_navigator_save_on_switch && @% != ""
       update
     endif
     let args = 'select-pane -' . tr(a:direction, 'phjkl', 'lLDUR')


plugin/tmux_navigator.vim
```diff
@@ -60,7 +60,8 @@ function! s:TmuxAwareNavigate(direction)
   " a) we're toggling between the last tmux pane;
   " b) we tried switching windows in vim but it didn't have effect.
   if tmux_last_pane || nr == winnr()
-    if g:tmux_navigator_save_on_switch
+    " Unly save the file if it has a filename
+    if g:tmux_navigator_save_on_switch && @% != ""
```
christoomey: Could you extract the non-file check to a function? Something like `s:isFileBacked()` or similar?

Also, is there any other hook that might be more explicit to check if the buffer is backed by a file? Something like `buftype`?

christoomey:
Hi @JackDanger, thanks for sharing this. I think this is a nice update. I made a few comments on the implementation that I was hoping you could work on, and then I'd be happy to merge this in.

keith:
Ping @JackDanger

plugin/tmux_navigator.vim
blueyed: Yes, this should use `&buftype` and `&modifiable` probably.

christoomey:
@blueyed Do you want to take this and finish out? Your comments re: `&bufname` and `&hidden` seem good. Consider it pre-approved and ready to merge assuming you don't find anything crazy.

paymog:
I think this was fixed with #143

christoomey:
@paymog I think you're right. I'm going to close this PR with that in mind, but all please feel free to reopen or comment if I missed something.

PR #95, opened 2015-11-12
Recognize binary name 'vimx'

patrickdepinguin:
Some distributions ship 'vim' without xterm_clipboard support. A separate
package vim-x11 or similar then ships another binary 'vimx' with that
support.

To support vim-tmux-navigator with such a binary name, adapt the pattern
check to allow an optional 'x'.


christoomey:
My kingdom for a more sane way to detect Vim. In the meantime, thanks for the update @patrickdepinguin!

PR #98, opened 2015-12-05
use * wildcard instead of regex matching to be more portable

Mischi:
Hi,

this makes vim-tmux-navigator work with OpenBSD's ksh(1) which
doesn't support regex matching.

I've found this solution on stackoverflow.
Please see http://stackoverflow.com/questions/229551/string-contains-in-bash/229585#229606

Thanks for this plugin!

Cheers,
Fabian


chris-tester-name:
Thanks for the update!

PR #104, opened 2016-01-25
Added functionality to check if Tmux is zoomed before passing through…

glentakahashi:
Not sure if this should be added as a flag, but I don't think many people like to switch panes when the current pane is maximized. This flag just check to see if the current vim process is in a maximized pane before sending the select-pane flag, so that tmux isn't unzoomed.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -42,6 +42,10 @@ function! s:TmuxCommand(args)
   return system(cmd)
 endfunction
 
+function! s:TmuxPaneIsZoomed()
+  return s:TmuxCommand("display-message -p '#{window_zoomed_flag}'") == 1
+endfunction
+
 function! s:TmuxPaneCurrentCommand()
   echo s:TmuxCommand("display-message -p '#{pane_current_command}'")
 endfunction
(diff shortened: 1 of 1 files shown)


christoomey:
Hi @glentakahashi, thanks for writing this up. Unfortunately, as it turns out this is functionality that I rely on on a daily basis. Specifically, my workflow is to run tests from vim (via [vim-tmux-runner](https://github.com/christoomey/vim-tmux-runner)), zoom the test running pane via tmux-runner, then use tmux-navigator to jump back (while zoomed). Tmux will automatically unzoom and I end up back in Vim where I left off. Very handy it turns out.

Based on the above I don't think we'll want to go forward with this PR, even behind a flag or option. I'm going to close this now with the above in mind, but please do reopen or comment if you think I misunderstood or you have questions.

glentakahashi:
Hm that's interesting that you use the feature like that. I've been active on some other threads relating to the same issue and I know that other people have been interested in disabling unzooming of tmux panes (unrelated to vim), so this functionality might apply to them as well. You can see the thread here: https://github.com/tmux/tmux/issues/150

I can see not having it as a default, but I think this functionality is very useful for people like me who compulsively press ctrl+j/ctrl+k and hate accidentally unzooming.

christoomey:
Hey @glentakahashi, thanks for adding the additional context. If I'm reading the thread correctly, it looks like they are heading towards making this behavior configurable in tmux. Assuming that is the case, then it would supersede any functionality in this plugin. The plugin is already more complex than I want it to be, so I'm very much keen to leave things like this out, especially when other tools can better handle the functionality. Apologies if this is less supportive of your workflow, I know that can be frustrating.

glentakahashi:
No problem, actually it's currently supported via running some custom commands, but that's beside the point. Thanks!

blueyed:
See #133 for a fresh approach at this.

PR #109, opened 2016-03-19
Add sane pane resizing to vim and tmux

arctelix:
Just some config that allows tmux panes and vim splits to be resized universally with Meta/Alt + [h,j,k,l].  It's like vim-tmux-navigator on STEROIDS, so i thought this was the best place for it to call home.  As a new vimer & tmuxer, i was shocked that apparently nobody else has done this??? Maybe not the best idea for some reason? but it feels so good...

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,30 @@ Thanks to Christopher Sexton who provided the updated tmux configuration in
 Configuration
 -------------
 
+### Add sane pane resizing to vim and tmux
+
+Allows for using Meta/Alt + [h,j,k,l] to resize tmux panes and vim splits.
+
+Add the following to your `~/.vimrc`:
+```
+" sane pane resizing
+nnoremap <silent> <M-k> <C-W>-
+nnoremap <silent> <M-j> <C-W>+
+nnoremap <silent> <M-l> <C-W>>
+nnoremap <silent> <M-h> <C-W><
+" optional auto equalize vim splits
+autocmd VimResized * wincmd =
+```
+
+Add the following to your `tmux.conf`:
+```
+# sane pane resizeing with awareness of vim splits
+bind -n M-j if-shell "$is_vim" "send-keys M-j" "resize-pane -D 10"
+bind -n M-k if-shell "$is_vim" "send-keys M-k" "resize-pane -U 10"
+bind -n M-l if-shell "$is_vim" "send-keys M-l" "resize-pane -R 10"
+bind -n M-h if-shell "$is_vim" "send-keys M-h" "resize-pane -L 10"
+```
+
 ### Custom Key Bindings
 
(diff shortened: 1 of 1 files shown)


christoomey:
Hey @arctelix, thanks for sharing! Personally I have different goals with regard to resizing windows in Vim, and panes in tmux. With Vim, I tend to go for a balanced window size (all equal via `wincmd =`, [with an an `autocmd` to automate](https://github.com/christoomey/dotfiles/blob/5c6189c90ab033788769ae6d277573404fc0c5e4/vim/rcfiles/visual#L87-L92)), whereas I have mappings for discreetly sizing tmux panes. Most often I have a Vim pane (75% of available space) and another "runner" pane for tests, shell, git, etc in the remaining 25%. That's just me though, I'm glad you were able to repurpose the VTN binding approach to serve your needs!

The main concern I have with this approach is I think it obscures a meaningful distinction by resizing the current thing. Sometimes I'll be in Vim (with multiple splits), but I want to resize the tmux pane Vim is in, not the current Vim window. I might be missing something, but I think that is a difference between pane navigation and window / pane sizing.

Beyond that, even though this plugin is pretty simple in operation, it ends up being surprisingly complex in implementation. With that in mind, I try to keep things as minimal and focused in the usage and documentation. As such, I'd rather not add these bindings to the README as I'm concerned it could confuse new users (even though I think they're great!).

With all that said, I'm going to close this issue, but I wanted to thank you again for sharing the work and trying to get it into the project. Frankly I think this would make a great blog post (and I'd be happy to link to that!) or similar, but I don't feel comfortable adding it to the already lengthy README. Please feel free to comment or reopen if you feel I've missed something or have additional questions.

arctelix:
Hey @christoomey, thanks for taking the time to share your methodology.  Only having four days under my belt with vim & tmux, your expert advice is totally apriciated.  Coming from intelij's PyCharm has spoiled me with easy pane and tab management, so this was a quick fix for me to get productive.  Most of vim & tmux concepts bindings make tons of sense and burned into my brain quickly.  The window management is what fried my Brian, having to pause for several seconds to sort out the proper command was horrible.

I do agree with your feelings about adding it to the README and figured as much.  So i'm going to put this and some other additions i have made into a gist.  I will post the link here when it's ready.

As to your comment about meaningful distinction, i agree there is some ambiguity and room for improvement with some internal detection, but this is a simple and great start.  Also, to be noted is that all of the default bindings are intact via <prefix> and <C-W>.  Plus easy access to sane tmux overrides, which you will see in my gist. 

I would greatly appreciate your feed back after reviewing the gist and thanks again!

arctelix:
Heres the gist, http://draft.sx/475cec89b6bba07c4d9a

Issue #110, opened 2016-03-21
Invalid arguments for TmuxWindCmd after vim update to 7.4 patch 1-1584.


phcerdan:
The plugin was working with no problems before vim update to 1-1584 in arch.
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Mar 17 2016 10:50:13)
Included patches: 1-1584
Compiled by Arch Linux

Using C-hjkl doesn't work inside vim and generates:
E116: Invalid arguments for function <SNR>79_TmuxWinCmd
Switching panes in tmux works.

phcerdan:
reverting to old version in arch solves* this.
`sudo pacman -U gvim-7.4.1529-1-x86_64.pkg.tar.xz vim-runtime-7.4.1529-1-x86_64.pkg.tar.xz`

ManiacalAce:
I used the [downgrade package](https://aur.archlinux.org/packages/downgrade/)  to downgrade to the above mentioned versions.

`downgrade vim-runtime gvim`

akemrir:
Same here. After downgrade plugin works.

saschagrunert:
Yep same here. Seems vim patch related.

nicolaiskogheim:
It didn't work right after downgrading for some reason, but after restarting Vim a couple of times it worked.

```bash
sudo pacman -U /var/cache/pacman/pkg/vim-7.4.1529-1-x86_64.pkg.tar.xz \
/var/cache/pacman/pkg/vim-runtime-7.4.1529-1-x86_64.pkg.tar.x
```
^ if you don't want to grab the `downgrade` plugin.

jck:
This seems to be fixed in 7.4.1639

jeanCarloMachado:
Solved here by 7.4.1639 as well.

djmaze:
Yes, can confirm it is fixed in 7.4.1639 (on Arch).

akemrir:
I confirm it too. Fixed.

christoomey:
Sounds like this is solved by the most recent patch (7.4.1639) so I'm going to close this. Thanks all for reporting and following up!

PR #112, opened 2016-03-23
Use `s:` instead of `<SID>` in commands

bullno1:
This addresses: https://github.com/christoomey/vim-tmux-navigator/issues/110
The use of `<SID>` is unnecessary in this context anyway.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -95,11 +95,11 @@ function! s:VimNavigate(direction)
   endtry
 endfunction
 
-command! TmuxNavigateLeft call <SID>TmuxWinCmd('h')
-command! TmuxNavigateDown call <SID>TmuxWinCmd('j')
-command! TmuxNavigateUp call <SID>TmuxWinCmd('k')
-command! TmuxNavigateRight call <SID>TmuxWinCmd('l')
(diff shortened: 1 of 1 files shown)


christoomey:
Hey @bullno1, thanks for the update. Can you explain a bit more about the distinction between `<sid>` and `s:`? I've always viewed them as largely interchangeable, but your PR leads me to believe this is not the case.

bullno1:
http://stackoverflow.com/questions/16768059/how-to-understand-these-vim-scripts

Basically, `s:` makes sense in a script context. It's used to address the local (sort of anonymous) namespace. `<SID>` is used in mapping, a global context. It gets substituted with whatever vim generated as the prefix for the script namespace. They used to be interchangeable but apparently recent vim patches decided to be pedantic about it.

blueyed:
:+1: in general.

But what about making `TmuxWinCmd` available as a public function?
So it can be accessed from outside.
Could also be moved to "autoload" while at it.. in a file `autoload/tmux_naviagor.vim` and would then be accessible as `tmux_navigator#tmuxwincmd` (bad name probably).

blueyed:
Thanks!
#131 came up as a duplicate, so it certainly helps as-is for now.

Issue #96, opened 2015-11-22
Use process list to determine if Vim is active


christoomey:
Currently we use a check of tmux's `pane_current_command` to determine if Vim is active. This has worked _just_ well enough, but is prone to a number of edge cases and false negatives.

Instead, I want to look into another approach that would examine the process tree for the pane, and detect Vim that way.

Via tmux's `{pane_pid}` format option to the `display-message` command we can get the parent pid of the pane. From there, we can use `pstree` to get the child process tree of the pane. The check is then if the current process (last element in `pstree` listing) is Vim.

A rough, half-working version of this is:

``` sh
ps -p $(tmux display-message -p -t 0 -F '#{pane_pid}' | xargs pstree | tail -1 | cut -d ' ' -f 3) -o comm=
```

That outputs `/usr/local/Cellar/macvim/7.4-73_1/MacVim.app/Contents/MacOS/Vim` in my case, which could be matched agains for Vim checking

Lost to improve here (is `pstree` available?, better `sed` for getting the leaf pid, etc), but it's a start I think

patrickdepinguin:
On my Linux system, the pstree output is separated with 3 dashes, for example:

    echo 4604 | xargs pstree 
    bash───xargs───vim

To get to vim, you can use the following command:

    echo 4604 | xargs pstree | tail -1 | rev | cut -d- -f1 | rev

The first echo would obviously need to be replaced.
Alternatively, since we're only interested in the last few characters being 'vim' (or derivative) or not, we could simply search for that string at the end, e.g.

    echo 4604 | xargs pstree | grep 'vim$'

blueyed:
> we could simply search for that string at the end

That won't work with nvim, which might have python subprocesses, and even childs inside (via `:terminal`):

  |   `-zsh,28494
  |       `-nvim,29542
  |           |-python3,29547 -c import sys; sys.path.remove(""); import neovim; neovim.start_host() script_host.py
  |           |-python3,29558 -c import sys; sys.path.remove(""); import neovim; neovim.start_host() /home/user/.vim/neobundles/deoplete/rplugin/python3/deoplete
  |           |-zsh,29577 -i
  |           |   `-nvim,29729
  |           |       |-pstree,29752 -apl
  |           |       |-python3,29734 -c import sys; sys.path.remove(""); import neovim; neovim.start_host() script_host.py
[...]

PR #117, opened 2016-04-08
Look for vim with ps/awk

blueyed:
This looks for PIDs using ps/awk, ignores a stopped Vim and its childs.

Based on the idea in #96.

TODO
 - [x] try to get rid of awk for performance reasons (`sed`/`cut` would be faster probably), or use a single `awk` invocation/script, which seems to be possible here (maybe also with `sed` even?!).

Leaving this here for reference, in case it might become necessary later:

    # Get last pid in list, without childs of Vim or stopped processes.
    # "T zsh" would be in there after (Neo)vim after "Ctrl-Z" and "fg".
    last_wo_vim_childs="$(echo "$session_childs" \
      | awk "\$2 != $vim_pid && \$1 != $$ && \$2 != $$ && \$3 !~ /T/ {print \$1}" \
      | tail -n1)"

    # We've filtered Vim's process group, so should end up at the pane_pid.
    test "$last_wo_vim_childs" -eq "$1" 2>/dev/null

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,14 +53,15 @@ Add the following to your `tmux.conf` file to configure the tmux side of
 this customization.
 
 ``` tmux
-# Smart pane switching with awareness of vim splits
+# Smart pane switching with awareness of Vim splits.
 # See: https://github.com/christoomey/vim-tmux-navigator
-is_vim='echo "#{pane_current_command}" | grep -iqE "(^|\/)g?(view|n?vim?x?)(diff)?$"'
-bind -n C-h if-shell "$is_vim" "send-keys C-h" "select-pane -L"
-bind -n C-j if-shell "$is_vim" "send-keys C-j" "select-pane -D"
-bind -n C-k if-shell "$is_vim" "send-keys C-k" "select-pane -U"
-bind -n C-l if-shell "$is_vim" "send-keys C-l" "select-pane -R"
-bind -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
+is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
+    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
+bind-key -n C-h if-shell "$is_vim" "send-keys C-h"  "select-pane -L"
+bind-key -n C-j if-shell "$is_vim" "send-keys C-j"  "select-pane -D"
+bind-key -n C-k if-shell "$is_vim" "send-keys C-k"  "select-pane -U"
+bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
+bind-key -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
 ```
 
 Thanks to Christopher Sexton who provided the updated tmux configuration in


README.md
```diff
@@ -53,14 +53,13 @@ Add the following to your `tmux.conf` file to configure the tmux side of
...
-# Smart pane switching with awareness of vim splits
+# Smart pane switching with awareness of Vim splits.
 # See: https://github.com/christoomey/vim-tmux-navigator
-is_vim='echo "#{pane_current_command}" | grep -iqE "(^|\/)g?(view|n?vim?x?)(diff)?$"'
-bind -n C-h if-shell "$is_vim" "send-keys C-h" "select-pane -L"
-bind -n C-j if-shell "$is_vim" "send-keys C-j" "select-pane -D"
-bind -n C-k if-shell "$is_vim" "send-keys C-k" "select-pane -U"
-bind -n C-l if-shell "$is_vim" "send-keys C-l" "select-pane -R"
-bind -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
+bind -n C-h if-shell "~/.dotfiles/vim/neobundles/tmux-navigator/is_vim.sh #{pane_pid}" "send-keys C-h" "select-pane -L"
```
blueyed: This could/should use `/path/to/is_vim.sh` probably?
It's shorter and needs to be changed in most cases anyway.

blueyed:
I've pushed some fixes.

keith:
Looks like this is incompatible with OS X because of differences with the `ps` flags.

blueyed:
@keith 
What is not supported? `-H`?

keith:
-H definitely. I didn't check the others.


--
Keith Smiley

blueyed:
@keith 
Given that the last part is now commented/disabled you could remove `-H` altogether.
It was used to get the processes ordered as a tree.
There would also be `f` for this, but that adds extra ASCII art.

keith:
Looks like `-s` doesn't exist on OS X as well.

keith:
Looked into this a little more and it seems like the `ps` and `pgrep` that ship with OS X are just super limited. It's possible to get the direct child pids of a process using `pgrep -P`, but you would have to do that multiple times to get down to the vim process. And with `ps` it seems like there's no way to do this.

Using `pstree` (`brew install pstree`) I was able to put together a small but annoying script to do what we want. Unfortunately it still required the use of `ps` to determine whether the process is suspended or not. Here's what I ended up with:

```sh
pids=$(pstree "$pid" | grep vim | cut -d" " -f5 | tr "\n" " ")
ps -o stat= -p $pids | grep S > /dev/null
```

blueyed:
@keith 
According to https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/ps.1.html, `-t` (for tty) should be available, too.
Please try the last update.

keith:
Works perfectly with `-t`! For reference the output is:

```
bash Ss
-zsh S
vim  S+
```

keith:
Definitely confirmed this works with OS X and in git owned processes that launch vim

blueyed:
Great.  I've squashed the commits and cleaned it up a bit more.

Should be ready then - I don't think the TODO is relevant really anymore, given that it's one `ps` and `awk` only by now.

blueyed:
@keith 
Can you check if `ps o state:1=,comm= -t "$1" | grep -qE "^[^T ]+ (\S+\/)?g?(view|n?vim?x?)(diff)?$"` also works for you?
(the important part being `ps o state:1=,comm=`)

christoomey:
@blueyed & @keith thanks for looking into this!

@blueyed, I get an error on OS X with the `ps o state:1=,comm=` invocation.

I'm having some luck with the following script though:

``` sh
#!/bin/sh

pid=$(tmux display-message -t "$1" -p '#{pane_pid}')
tty=$(ps -p "$pid" | tail -1 | cut -d ' ' -f 3)
process=$(ps -t "$tty" | tail -n +2 | cut -d ' ' -f 8 | tail -1)

echo "$process" |  grep -iqE "(^|\/)g?(view|n?vim?x?)(diff)?$"
```

Some inconsistency between output that I'm battling, but heading in the right direction. I'll poke more later this week.

keith:
I also have an error with the `state:1=`. Is there an issue with the script in the current diff?

blueyed:
The current diff/script should be fine, but can be optimized.
See the fixup commit I've just pushed.
It's just `ps` and `grep` now.

blueyed:
@christoomey 
Have you tried this again?

It seems to work really well for me.

I'm not sure though anymore that a separate script makes sense, given that it's only `ps … | grep …` in the end.

christoomey:
I've tried it and it seems to be really close, but still slightly off for me. I'll poke more soon and get back to you with specifics. Feels like we're very close.

I would love to move the script back inline now that it is a bit more concise.

blueyed:
> I would love to move the script back inline now that it is a bit more concise.

Given https://github.com/christoomey/vim-tmux-navigator/pull/123 I think it makes sense to have it in a include file, and refer to that from the readme, right?

blueyed:
@christoomey 
> but still slightly off for me

For debugging, try adding `set -x; exec 2>>/tmp/debug.log` and then tailing `/tmp/debug.log` in a tmux pane.

raviqqe:
@blueyed 
Is't the code block changed in `README.md` a piece of tmux script?
You can't `source` the shell script, `vim-tmux-navigator.tmux` there.

blueyed:
Yeah, should be run-shell.

blueyed:
Fixed.

raviqqe:
I tested @blueyed's branch with the 4 combinations of {TPM, run-shell} x {Ubuntu 14.04, FreeBSD 10.3} manually.
And, they all work well!

christoomey:
The current version of the script doesn't work for me, but I was able to tweak it to get it work consistently for both normal Vim as primary process, and Vim started from within git.

```
is_vim="ps o state=,comm= -t '#{pane_tty}' \
    | tail -1 | awk '{print $2}' | grep -iqE '(^|\/)g?(view|n?vim?)(diff)?$'"
```

Thoughts?

blueyed:
What's the output of `ps o state=,comm= -t '#{pane_tty}'` when it fails?
You may want to use a separate script for testing/debugging (with `set -x`, some echos and redirecting to a file).  But you can also run the `ps` with `-t` from another terminal emulator of course.
Your version won't handle stopped Vim processes probably.

raviqqe:
My manual checks were wrong. I have totally missed the point.
So, I rerun the new test and it passed on Ubuntu 14.04. However, on FreeBSD 10.3, it failed.
This is because its `ps` command's `o` option takes an equal sign only at the end of the last keyword and a whole string following it becomes a new column name.
I mean:

```
$ ps o state=,comm= # ",comm=" is the new column name.
,comm=
Ss
R+
```

My solution is:

```
$ ps -o state= -o comm= # This runs on both Ubuntu and FreeBSD!
Ss zsh
R+ ps
```

Could you adapt it to support FreeBSD (my favorite OS)?
Thanks!

blueyed:
Done in 4664981.
Thanks for debugging this.

@christoomey 
That might also fix your issue since MacOS's `ps` might behave the same.

blueyed:
@christoomey 
Please confirm that it works for you now, so that this can be merged.

keith:
+1 this is a huge improvement! // @christoomey

blueyed:
Pushed another small update, especially regarding the commit message.

Will merge this shortly, assuming that @christoomey does not find the time to re-check it currently: it has been tested by others (also on MacOS).

As a gimmick: I could use the same mechanism for my auto-PageUp tmux binding, which did not work when using `less --no-init` (to not use an alternate screen with `less`): https://github.com/blueyed/dotfiles/commit/43e99d510d97254edb470fed2a005f8567e8d19a

christoomey:
@blueyed thanks for continued effort to push this forward, and apologies for slowing this up with lack of responsiveness.

A couple of things I'd like to see before we merge this:

1. I think the `grep` needs to be updated to inlcude `-i`. This is required for my usecase (MacVim Vim which shows up as `/usr/local/Cellar/macvim/7.4-73_1/MacVim.app/Contents/MacOS/Vim` for me)
2. I'd like to split out the work needed to support TPM (and with that I'd still prefer to have the README text still work without needing to require the external script).
3. Unfortunately this is failing for me even with the addition of `-i` on the grep. Not sure why but will need to debug.

For number 3 above, it's odd as running the commands directly as below does seem to work, but Vim detection as part of key mappings fails.

This works at the command line as expected:

``` sh
ps -o state= -o comm= -t "$(tmux display-message -t 2 -p "#{pane_tty}")" \
    | grep -iqE '^[^TXZ ]+ +(\S+\/)?g?(view|n?vim?x?)(diff)?$'
```

rosshadden:
This works very well for me on Arch Linux, using `termite`.  What you're doing with `ps` is brilliant, and is the closest we've ever been to a perfect solution.

It does not work in vim within ssh sessions.  I'm not sure if it should, nor even how it could, but I think it's worth mentioning.

It does not work in vim in a nested tmux session, with one running locally and one running on a remote host.  You would think if the above were true, it would be obvious for this to be as well.  But with my prior config, this did work (even though navigating through terminals without vim opened did not.

```sh
isVim="tmux display -p '#{pane_current_command} #T' | grep -qE ' - N?VIM$'"
```

Edge cases, to be sure.  Since they may well only effect me, I will dig into them myself, and they should not necessarily slow progress on this pull request.

rosshadden:
Just an update to my above post.  It's actually strange that 

```sh
ps -o state= -o comm= -t "$(tmux display-message -t 2 -p "#{pane_tty}")" | grep -iqE '^[^TXZ ]+ +(\S+\/)?g?(view|n?vim?x?)(diff)?$'
```

doesn't work inside `ssh`, because the output of the command is exactly the same as it is locally:

```sh
S zsh
S nvim
```

I will continue to look into this, but it might be related to why this doesn't work for @christoomey.

blueyed:
@christoomey 
Thanks!

1. I'll add the `-i`
2. I thought it was a win-win: it's better to have it in a separate script (where you can copy'n'paste from), and is required for tpm anyway: therefore it should be only defined in one place.
We can have it in the README still verbatim though.  I can see that it's easier to setup that way.
I'll also add it to the README.
3. Maybe it's related to quoting?!  I suggest trying to use an external script and then use `set -x` and redirection to a file to see what's going on.

README.md
```diff
@@ -53,14 +53,23 @@ Add the following to your `tmux.conf` file to configure the tmux side of
...
-bind -n C-l if-shell "$is_vim" "send-keys C-l" "select-pane -R"
-bind -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
+
+has_vim="ps -o state= -o comm= -t '#{pane_tty}' \
+    | grep -iEq '^[^TXZ ]+ +(\S+\/)?g?(view|n?vim?x?)(diff)?$'"
+tmux bind-key -n C-h  if-shell "$has_vim" "send-keys C-h"  "select-pane -L"
+tmux bind-key -n C-j  if-shell "$has_vim" "send-keys C-j"  "select-pane -D"
+tmux bind-key -n C-k  if-shell "$has_vim" "send-keys C-k"  "select-pane -U"
+tmux bind-key -n C-l  if-shell "$has_vim" "send-keys C-l"  "select-pane -R"
+tmux bind-key -n C-\\ if-shell "$has_vim" "send-keys C-\\" "select-pane -l"
```
christoomey: I was able to sort this out. Looks like tmux escaping the format string (specifically the `\S+\/` was being stripped down to `S+/`). The final form should be:

``` tmux
is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
bind-key -n C-h  if-shell "$is_vim" "send-keys C-h"  "select-pane -L"
bind-key -n C-j  if-shell "$is_vim" "send-keys C-j"  "select-pane -D"
bind-key -n C-k  if-shell "$is_vim" "send-keys C-k"  "select-pane -U"
bind-key -n C-l  if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
bind-key -n C-\\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
```

Note, this also removes likely accidental additions of `tmux` prefixing each command (likely a carryover from TPM config?).

Also, I'd like to stick to the `is_vim` nomenclature as I think it better captures the intent (especially with the addition of support for a backgrounded Vim session thanks to you @blueyed), and I'd prefer not to change that if we don't have a strong reason.

README.md
```diff
@@ -53,14 +53,23 @@ Add the following to your `tmux.conf` file to configure the tmux side of
...
+```
+
+The above code snippet is also available in the `vim-tmx-navigator.tmux` file,
+which you could alternatively include like this in your tmux config:
+
+```tmux
+run-shell /path/to/vim-tmux-navigator/vim-tmux-navigator.tmux
```
christoomey: I'd prefer to remove this and the associated file entirely from this PR, in favor of bringing those changes in later via the TPM configuration in works in #123.

vim-tmux-navigator.tmux
```diff
@@ -0,0 +1,13 @@
...
+# tpm (https://github.com/tmux-plugins/tpm).
+
+has_vim="ps -o state= -o comm= -t '#{pane_tty}' \
+    | grep -iEq '^[^TXZ ]+ +(\S+\/)?g?(view|n?vim?x?)(diff)?$'"
+
+tmux bind-key -n C-h  if-shell "$has_vim" "send-keys C-h"  "select-pane -L"
+tmux bind-key -n C-j  if-shell "$has_vim" "send-keys C-j"  "select-pane -D"
+tmux bind-key -n C-k  if-shell "$has_vim" "send-keys C-k"  "select-pane -U"
+tmux bind-key -n C-l  if-shell "$has_vim" "send-keys C-l"  "select-pane -R"
+tmux bind-key -n C-\\ if-shell "$has_vim" "send-keys C-\\" "select-pane -l"
```
christoomey: Let's remove this in favor of merging the TPM config in #123 after we get this update in.

christoomey:
@blueyed I think we've finally got a working version! Again, thanks a ton for continued efforts to move this forward.

I've commented on a number of things inline, but I think we're very close to having a mergeable version. Would you mind doing the following to get us ready?:

1. Test out the provided snippet (using the inline form in your ~/.tmux.conf) to confirm all stick works for you.
2. Assuming it's good, update the README section as outlined
3. Remove the external script and notes about that. The TPM config in #123 will cover that, and I would like to keep this PR as focused as possible given the scope of it.

Once we have all that in place, I think we should be good.

For added sanity check, @keith & @geoffharcourt could you please update your config locally with the new code (included below) and confirm all still works for you (including Vim sessions started as children of other processes, like git!)

``` tmux
is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
bind-key -n C-h  if-shell "$is_vim" "send-keys C-h"  "select-pane -L"
bind-key -n C-j  if-shell "$is_vim" "send-keys C-j"  "select-pane -D"
bind-key -n C-k  if-shell "$is_vim" "send-keys C-k"  "select-pane -U"
bind-key -n C-l  if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
bind-key -n C-\\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
```

rosshadden:
The updated conditional works great for me.

blueyed:
@christoomey 
`bind-key -n C-\\` does not work for me inline in tmux: `Unknown key: C-\\`?!

I've just used `C-\` there now.

blueyed:
Rebased/squashed it.

Hopefully comment number 42 will be a "go".. ;)

christoomey:
All that effort for such a tiny diff! @blueyed, we're a go. Merge away.

christoomey:
👍  I think the expanded `bind-key` is an improvement. Thanks again for keeping this moving. 🚢 away!

blueyed:
:rocket: 

You're welcome.. :)

keith:
Awesome work!

blueyed:
Indeed.
Thank you all for the good feedback! :)

blueyed:
FWIW: submitted the plugin itself to reddit: https://www.reddit.com/submit?url=https%3A%2F%2Fgithub.com%2Fchristoomey%2Fvim-tmux-navigator%2F

nicola:
Vim to tmux doesn't work for me on the mac :(

blueyed:
@nicola 
Please file a new issue for that and provide more information, e.g. how the snippet from the README works (e.g. what the output of `ps ..` looks like when using the target (vim) pane's tty).

PR #123, opened 2016-04-30
Support TPM (Tmux Plugin Manager)

raviqqe:
Support tmux configuration via TPM (Tmux Plugin Manager).
As tmux allows overriding bound keys, this seems to be
just a convenient feature for both newbies and experienced users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,14 @@ bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
 bind-key -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
 ```
 
+Or optionally, when using the [TPM][] add the following lines to
+`~/.tmux.conf`. For more information see [TPM][].
+
+``` tmux
+set -g @plugin 'christoomey/vim-tmux-navigator'
+run '~/.tmux/plugins/tpm/tpm'
+```
+
 Thanks to Christopher Sexton who provided the updated tmux configuration in
 [this blog post][].
 
@@ -236,6 +244,7 @@ script][] which has a more robust check.
 [Mislav Marohnić's]: http://mislav.uniqpath.com/
 [Mislav's original external script]: https://github.com/mislav/dotfiles/blob/master/bin/tmux-vim-select-pane
 [Vundle]: https://github.com/gmarik/vundle
+[TPM]: https://github.com/tmux-plugins/tpm
 [configuration section below]: #custom-key-bindings
 [this blog post]: http://www.codeography.com/2013/06/19/navigating-vim-and-tmux-splits
 [this gist]: https://gist.github.com/mislav/5189704
--- a/vim-tmux-navigator.tmux
+++ b/vim-tmux-navigator.tmux
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
+    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
+tmux bind-key -n C-h if-shell "$is_vim" "send-keys C-h"  "select-pane -L"
+tmux bind-key -n C-j if-shell "$is_vim" "send-keys C-j"  "select-pane -D"
+tmux bind-key -n C-k if-shell "$is_vim" "send-keys C-k"  "select-pane -U"
+tmux bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
+tmux bind-key -n C-\\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"


blueyed:
Makes sense, thanks!

But we should probably merge https://github.com/christoomey/vim-tmux-navigator/pull/117 first, which changes the config snippet to be included (which gets done through ` vim-tmux-navigator.tmux`, correct?).

raviqqe:
I totally agree.
So, after that, we need to rebase the commits in this PR.

blueyed:
For the README consider adding something like this (via https://github.com/christoomey/vim-tmux-navigator/pull/117#discussion_r65722678 / https://github.com/christoomey/vim-tmux-navigator/pull/117/commits/0ae33100d5668aca42c88237e69e832174456574):

    The above code snippet is also available in the `vim-tmx-navigator.tmux` file,
    which you could alternatively include like this in your tmux config:

    ```tmux
    run-shell /path/to/vim-tmux-navigator/vim-tmux-navigator.tmux
    ```

blueyed:
Just wanted to leave the above for completeness - feel free to ignore (most of) it.. :)

blueyed:
#117 is finally merged, please rebase.

raviqqe:
The new commits are tested only on FreeBSD now.

By the way, do these commits follow your intention, everyone?
As discussed in #117, there are 2 ways to realize this ("define in one place" vs. "duplicating script").
These commits picked the latter.

README.md
```diff
@@ -64,6 +64,15 @@ bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
 bind-key -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
 ```
 
+Or optionally, use [TPM][] (tmux plugin manager) installing it first
+and adding the lines below to `~/.tmux.conf`.
```
blueyed: better: "Or optionally, when using the [TPM][] (tmux plugin manager) add the following lines to ..."

Not really sure, but the current version is bumpy ("installing it first").

README.md
```diff
@@ -64,6 +64,15 @@ bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
...
+For more information, see [TPM][].
```
blueyed: Needs no comma.

raviqqe:
Thank you for your flattening bumps!
Any others?

vim-tmux-navigator.tmux
```diff
@@ -0,0 +1,13 @@
...
+  local is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
+      | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
+  tmux bind-key -n C-h if-shell "$is_vim" "send-keys C-h"  "select-pane -L"
+  tmux bind-key -n C-j if-shell "$is_vim" "send-keys C-j"  "select-pane -D"
+  tmux bind-key -n C-k if-shell "$is_vim" "send-keys C-k"  "select-pane -U"
+  tmux bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
+  tmux bind-key -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
+}
+
+bind_keys
```
christoomey: For my personal knowledge, can you explain why the wrapper function is used and then immediately invoked? Is it intended to prevent leaking the `is_vim`?

raviqqe:
Yes, it is. It's for not dirtying the global scope and a little bit of abstraction.
However, it would be more straightforward and simpler without the function definition.
I'm not sure which is better.

vim-tmux-navigator.tmux
raviqqe: Yes, it is. It's for not dirtying the global scope and a little bit of abstraction.
However, it would be more straightforward and simpler without the function definition.
I'm not sure which is better.

blueyed: Will the file be sourced or run?
Only when sourcing it the var would leak.

raviqqe: According to [this file](https://github.com/tmux-plugins/tpm/blob/master/scripts/source_plugins.sh) of TPM, they are run.
So, I will flatten `bind_keys`.

raviqqe:
Now, its flattened. The script is tested on FreeBSD 10.3 and Fedora 23.

raviqqe:
Now, its flattened and tested on FreeBSD 10.3 and Fedora 23.

blueyed:
You've forgotten something more from my patch.. (the paragraph about `run-shell`).

(I've just tested this to make sure that `$is_vim` does not leak there)

What about the following patch on top of this PR?

    diff --git i/README.md w/README.md
    index 68b4e36..8bd06a8 100644
    --- i/README.md
    +++ w/README.md
    @@ -47,10 +47,13 @@ Then run
     :PluginInstall
     ```
     
    -### Tmux
    +### tmux
     
    -Add the following to your `tmux.conf` file to configure the tmux side of
    -this customization.
    +To configure the tmux side of this customization there are three options:
    +
    +#### Add a snippet
    +
    +Add the following to your `tmux.conf` file:
     
     ``` tmux
     # Smart pane switching with awareness of Vim splits.
    @@ -64,7 +67,18 @@ bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
     bind-key -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
     ```
     
    -Or optionally, when using the [TPM][] add the following lines to
    +#### Include the snippet
    +
    +The above code snippet is also available in the `vim-tmux-navigator.tmux`
    +file, which can be includes like this in your tmux config:
    +
    +```tmux
    +```
    +
    +#### TPM
    +
    +When using Tmux Plugin Manager ([TPM][]) add the following lines to
     `~/.tmux.conf`. For more information see [TPM][].
     
     ``` tmux

I suggest squashing this into your commit and then merging it - so there's no need for you to update the PR once more.

blueyed:
Closing this in favor of #132.

PR #126, opened 2016-05-10
Added a method of retaining C-l clear screen for full-width panes

MarSoft:


README.md
```diff
@@ -121,6 +121,16 @@ With this enabled you can use `<prefix> C-l` to clear the screen.
...
 Thanks to [Brian Hogan][] for the tip on how to re-map the clear screen binding.
 
+#### Intellectual Clear Screen
+
+Additionally, it is possible to make `<Ctrl-l>` to work as usual if current pane is full-width (this will also work well if that pane is Vim):
+
+``` tmux
+single_pane='test $(tmux list-panes | grep "(active)" | grep -o "\[[0-9]*x" | tr -d "[x") -eq $(tmux list-windows | grep "(active)" | grep -o " \[[0-9]*x" | tr -d "[x")'
+...
+bind -n C-l if-shell "$single_pane || $is_vim" "send-keys C-l" "select-pane -R"
```
blueyed: Nice, but it can be made easier:

`bind -n C-l if '[ #{pane_width} = #{window_width} ]' "send-keys C-l" "select-pane -R"`

(This also includes `window_zoomed_flag==1`.)

blueyed:
I'd say to wait for #117, but then it should probably not be the default?!

Other opinions?

README.md
MarSoft: Great, thanks! (I am a newbie with tmux..)
11 мая 2016 г. 0:14 пользователь "Daniel Hahler" <notifications@github.com>
написал:

christoomey:
Hey @MarSoft, thanks for sharing this and apologies for my delay in replying. While this is a novel solution, I'm very wary of adding any additional complexity to the setup. In addition, I think overloading `C-l` to do clear screen in addition to the already complex navigation is more than I'm comfortable bringing into the plugin. For my use case I've found the consistency of the `prefix C-l` mapping to work great and now require any thought.

Again, apologies for the delayed response and I do hope this doesn't discourage you from potential future contributions. As I said, I think this is a very novel and interesting solution, I just tend to err on the side of simplicity and consistency for this plugin.

PR #132, opened 2016-06-15
Followup to #123: Support TPM (Tmux Plugin Manager)

blueyed:
See #123 for the discussion.  This adds an improvement to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,14 @@ bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
 bind-key -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
 ```
 
+Or optionally, when using the [TPM][] add the following lines to
+`~/.tmux.conf`. For more information see [TPM][].
+
+``` tmux
+set -g @plugin 'christoomey/vim-tmux-navigator'
+run '~/.tmux/plugins/tpm/tpm'
+```
+
 Thanks to Christopher Sexton who provided the updated tmux configuration in
 [this blog post][].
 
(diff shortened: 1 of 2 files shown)


raviqqe:
Thank you, @blueyed for your taking over and continuing work.
Is there any problem with this branch now? Validation needed?

blueyed:
@raviqqe 
It's ready from my side.

@christoomey what do you think?

README.md
```diff
@@ -64,6 +67,25 @@ bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
...
 ```
 
+#### Include the snippet
+
+The above code snippet is also available in the `vim-tmux-navigator.tmux`
+file, which can be includes like this in your tmux config:
+
+```tmux
+```
```
christoomey: I think this section is potentially confusing and I'd rather remove it altogether. While I appreciate the desire to make install easier, I've in general opted for clarity and directness rather than fully enumerating all the possibilities. With this section we're offering 3 different ways to set up. I'd prefer to remove this and clarify the other two to make this all clearer.

README.md
```diff
@@ -64,6 +67,25 @@ bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
...
+
+#### TPM
+
+When using Tmux Plugin Manager ([TPM][]) add the following lines to
+`~/.tmux.conf`. For more information see [TPM][].
```
christoomey: Can we change the language here to make it more clear that this is an alternative to the "copy this snippet into your tmux.conf file" section? Something like:

> If you'd prefer, you can use the Tmux Plugin Manager ([TPM][]) instead of copying the snippet.
> When using TPM, add the following lines to your `~/.tmux.conf`

README.md
```diff
@@ -47,10 +47,13 @@ Then run
...
-### Tmux
+### tmux
 
-Add the following to your `tmux.conf` file to configure the tmux side of
-this customization.
+To configure the tmux side of this customization there are three options:
+
+#### Add a snippet
+
+Add the following to your `tmux.conf` file:
```
christoomey: Can you make this `~/.tmux.conf` to match the other references in the file?

christoomey:
I added a few comments inline about some of the wording, and suggesting we remove the "or source this snippet with..." section. Beyond that, I think this looks great and will be a nice addition / simplification to the install process for TPM folks. Thanks @raviqqe for kicking off, @blueyed for keeping it going. Feel free to squash and merge after addressing my comments 👍

blueyed:
Okay, I've amended my commit.
Please press the button if you think it's fine.. :)

christoomey:
This has been merged into master as d5dc8d3c80cf898a25b8cf16ab65cf6a5438009f.

Issue #56, opened 2014-12-30
About Tmux zoom feature


oblitum:
I'd like to put in discussion what would be the best behavior when you zoom a particular pane. For me it's a bit annoying when I'm in a zoomed pane and get out of zoom by accident when moving around buffers/panes.

Would it be possible/desirable to have vim-tmux-navigation  disabled (both for tmux as for vim) when zoom is ON in a given pane? Because I find vim-tmux-navigation useful solely when zoom is off.

oblitum:
Just wanted to report here that, form some reason, recently on Archlinux with an up to date tmux and vim-tmux-navigator, the plugin is behaving the way I wished. I'm forced to zoom out to tmux-navigate.

blueyed:
See #133 for a fresh approach at this.

oblitum:
@blueyed thanks for letting me know... current behaviour for me is back as formerly explained in the issue, it seems I've learned to live with it, but if you got a fix I'll sure switch, I don't like the default.

blueyed:
Yeah, try #133.  It even let's you zoom out on 2nd invocation by default.
I think that's what @christoomey could live with even - but there's also a setting to get the current behavior.

oblitum:
Just trying.

PR #133, opened 2016-06-15
Improve moving out of zoomed tmux pane

blueyed:
There were issues (e.g. #56) and pull requests (#65 and #104) to improve this
behavior, but they were mainly rejected because it breaks other use cases.

This PR adds an option that can be enabled to have the current behavior
(`let g:tmux_navigator_zoomed_tmux_delay = 0`), but it also
recognizes when the user tries to get out of the zoomed pane twice, and
then allows for it, using a configurable delay (`let g:tmux_navigator_zoomed_tmux_delay = 0.5` - the default).
`g:tmux_navigator_zoomed_tmux_delay = -1` can be used to display a tmux message, but not zoom out.

TODO:
- [ ] doc

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -82,22 +82,25 @@ function! s:TmuxAwareNavigate(direction)
 
     " Handle delay for moving out of zoomed tmux pane.
     " This wraps the tmux args conditionally.
+    " 0 to zoom out always. -1 can be used to never zoom out.
     let timeout = get(g:, 'tmux_navigator_zoomed_tmux_delay', 0.5)
     if timeout != 0
       let msg = 'Tmux is zoomed, not moving out.'
-      let curtime = reltimefloat(reltime())
-      if (a:direction != s:tried_moving_out_of_zoomed_tmux[0]
-            \ || curtime > (s:tried_moving_out_of_zoomed_tmux[1] + timeout))
-        " 0 to zoom out always
-        " -1 can be used to never zoom out.
-        let msg .= printf(' Trigger it again in %.1f seconds to zoom out.',
-              \ timeout)
-        let msg .= ' See the help for g:tmux_navigator_zoomed_tmux_delay.'
-        let s:tried_moving_out_of_zoomed_tmux = [a:direction, curtime]
+      if timeout > 0
+        let curtime = reltimefloat(reltime())
+        if (a:direction != s:tried_moving_out_of_zoomed_tmux[0]
+              \ || curtime > (s:tried_moving_out_of_zoomed_tmux[1] + timeout))
+          let msg .= printf(' Trigger it again in %.1f seconds to zoom out.',
+                \ timeout)
+          let msg .= ' See the help for g:tmux_navigator_zoomed_tmux_delay.'
+          let s:tried_moving_out_of_zoomed_tmux = [a:direction, curtime]
+        else
+          let s:tried_moving_out_of_zoomed_tmux = ['', 0]
+        endif
+      endif
+      if timeout < 0 || s:tried_moving_out_of_zoomed_tmux[1] != 0
         let args = "if -F '#{window_zoomed_flag}' 'display \"".msg."\"'"
               \ ." '".args."'"
-      else
-        let s:tried_moving_out_of_zoomed_tmux = ['', 0]
       endif
     endif
     silent call s:TmuxCommand(args)
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -10,6 +10,9 @@ let g:loaded_tmux_navigator = 1
 if !exists("g:tmux_navigator_save_on_switch")
   let g:tmux_navigator_save_on_switch = 0
 endif
+if !exists('g:tmux_navigator_zoomed_tmux_delay')
+  let g:tmux_navigator_zoomed_tmux_delay = 0.5
+endif
 
 function! s:TmuxOrTmateExecutable()
   return (match($TMUX, 'tmate') != -1 ? 'tmate' : 'tmux')
@@ -83,7 +86,7 @@ function! s:TmuxAwareNavigate(direction)
     " Handle delay for moving out of zoomed tmux pane.
     " This wraps the tmux args conditionally.
     " 0 to zoom out always. -1 can be used to never zoom out.
-    let timeout = get(g:, 'tmux_navigator_zoomed_tmux_delay', 0.5)
+    let timeout = g:tmux_navigator_zoomed_tmux_delay
     if timeout != 0
       let msg = 'Tmux is zoomed, not moving out.'
       if timeout > 0
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -79,6 +79,27 @@ function! s:TmuxAwareNavigate(direction)
       endtry
     endif
     let args = 'select-pane -t ' . shellescape($TMUX_PANE) . ' -' . tr(a:direction, 'phjkl', 'lLDUR')
+
+    " Handle delay for moving out of zoomed tmux pane.
+    " This wraps the tmux args conditionally.
+    let timeout = get(g:, 'tmux_navigator_zoomed_tmux_delay', 0.5)
+    if timeout != 0
+      let msg = 'Tmux is zoomed, not moving out.'
+      let curtime = reltimefloat(reltime())
+      if (a:direction != s:tried_moving_out_of_zoomed_tmux[0]
+            \ || curtime > (s:tried_moving_out_of_zoomed_tmux[1] + timeout))
+        " 0 to zoom out always
+        " -1 can be used to never zoom out.
+        let msg .= printf(' Trigger it again in %.1f seconds to zoom out.',
+              \ timeout)
+        let msg .= ' See the help for g:tmux_navigator_zoomed_tmux_delay.'
+        let s:tried_moving_out_of_zoomed_tmux = [a:direction, curtime]
+        let args = "if -F '#{window_zoomed_flag}' 'display \"".msg."\"'"
+              \ ." '".args."'"
+      else
+        let s:tried_moving_out_of_zoomed_tmux = ['', 0]
+      endif
+    endif
     silent call s:TmuxCommand(args)
     if s:NeedsVitalityRedraw()
       redraw!
@@ -88,6 +109,7 @@ function! s:TmuxAwareNavigate(direction)
     let s:tmux_is_last_pane = 0
   endif
 endfunction
+let s:tried_moving_out_of_zoomed_tmux = ['', 0]
 
 function! s:VimNavigate(direction)
   try


oblitum:
I think the behaviour is a bit inconsistent for me. It's providing a warning at the very first time I hit the edge, but then, if I hit it again it zooms out and never brings back the warning anymore when I zoom back the pane with Vim.

oblitum:
It would be nice also if you multiplexed this option for:

- default
- warn on first hit and zoom out and switch on second hit
- never zoom out switch panes, just warn.

blueyed:
@oblitum 
Yeah, you have to move inside of Vim again to re-trigger it..
I've had a separate variable/state for it in the beginning, but then thought that `tmux_is_last_pane` could be re-used.

It also does not work for me when moving out of a Vim to the right or down..

Let's assume for now that this can be fixed and the behavior is like when moving up/left for the first time (or after moving inside of Vim).  Do you think it could fly?  /cc @christoomey

oblitum:
It's not working for me in any direction after I've moved outside Vim the first time. I get the expected behaviour only at the first time I move outside Vim, when I get back and zoom it again I don't get the warning anymore, just the original behaviour.

I think it's a nice scheme if this gets fixed and works, in any directions, consistently, maybe turning it the default one? Since it's just in the middle of the extreme behaviours I listed.

oblitum:
by the way, it's working for me (in the first time), when moving down (also right).

oblitum:
What I'm considering an issue may be a clash in expectations. I was expecting that this added behaviour of zooming out and switching on the second hit would happen on "insistence", meaning, after hitting a second time in a short period of time.

blueyed:
@oblitum 
Good idea making it time based.
I've changed that.  It uses 2-3 seconds as a timeout currently.
Should it be as long as 5?

oblitum:
@blueyed for me current timing is good.

oblitum:
@blueyed do you have an opinion on https://github.com/christoomey/vim-tmux-navigator/pull/133#issuecomment-226636435?

oblitum:
I've noticed that, when zoomed, it's working only when there's a hidden open pane in the direction I try to move. If there's no pane in that direction, but there's another in other direction,it zooms out. So, if I split horizontally and zoom the top panel and try to move down, it warns as expected, if I try to move right it just zooms out, because there's no pane there.

blueyed:
I like that set of options.
Can it be made with a `delay` option, where -1 would be the current behavior, 0 warns all the time and otherwise it's the number of seconds for the re-triggering? (i.e. 2)

oblitum:
> I like that set of options.
Can it be made with a delay option, where -1 would be the current behavior, 0 warns all the time and otherwise it's the number of seconds for the re-triggering? (i.e. 2)

Looks nice.

oblitum:
I mean, looks nice and can also be the sole option controlling this feature, removing the former boolean one.

blueyed:
Added that.

Are you sure about your previous comment that it does not work?
Seems to work for me in all cases now?!

We should probably also consider the direction with the timeout?

oblitum:
> > Added that.

> I guess you mean locally? Didn't got the update.

Ah, it's `g:tmux_navigator_move_out_of_zoomed_tmux`. Ok then.

blueyed:
Yeah, should be added to the doc of course once it's settled.

oblitum:
Ok, I've tried to test the behavior we agreed but it's not as stated. I've checked the code too, it seems the `let g:tmux_navigator_move_out_of_zoomed_tmux = 0` is not being covered as expected, if I try to hit fast it warns and switches, `0` was expected to not warn just switch like original behavior. `-1` is working like original behavior, but we agreed it to mean never switch.

blueyed:
Fixed in 0fae03b hopefully?!

blueyed:
No, scratch that.

oblitum:
@blueyed From the previous behaviour I was inferring that `get(g:, 'tmux_navigator_move_out_of_zoomed_tmux', 2)` was giving 2 when `g:tmux_navigator_move_out_of_zoomed_tmux` was `0`. I guess you need to check for variable existence instead, to set the default value. If I'm guessing right this is the culprit.

blueyed:
Well, I wrote something different initially.

> -1 would be the current behavior, 0 warns all the time and otherwise it's the number of seconds for the re-triggering?

So it was just -1 and 0 switched?!
See my last fixup, which also changes the msg dynamically.
It is easier to use -1 for "warn, but not switch" and 0 for the current behavior (no delay, just zoom out).

`get(g:, 'foo', 2)` will be 0 for foo=0.

oblitum:
@blueyed Great! Yeah we agreed reversed formally, but for me it's fine now, `0` meaning no-delay, `-1` blocking.

So, feature wise, I guess it's all done! I really hope at last this tentative get merged, much better options now.

oblitum:
@blueyed ah, there's still a gotcha! I think this is the last one.

I've put `5` in my `.vimrc`, and, when I hit twice for zooming out and switching, if I go back many times before 5 seconds passes and zoom in, it zooms out without notice because the time was not reset on the first switch.

oblitum:
So, while the timer doesn't timeout, it operates in original behaviour mode. I guess it should be reset for this not to happen.

blueyed:
Cool.
We could probably reset the timer or rather the marker altogether (which also holds the direction) on FocusGained or rather VimResized.  Will not do this now, but I agree that it should be done.
Thanks for testing this out and the valuable feedback! :)

oblitum:
@blueyed I think it's better for the code to be refactored locally instead of spreading the patch around. For example, the following fix the problem:

```vim
    let timeout = get(g:, 'tmux_navigator_move_out_of_zoomed_tmux', 2)
    let timedout = (localtime() > s:tried_moving_out_of_zoomed_tmux[1] + timeout)
    if timeout != 0 && (a:direction != s:tried_moving_out_of_zoomed_tmux[0] || timedout)
      let msg = "Tmux is zoomed, not moving out."
      if timeout > 0
        let msg .= 'Trigger it again, or set g:tmux_navigator_move_out_of_zoomed_tmux=0 to zoom out always.  -1 can be used to never zoom out.'
      endif
      let args = "if -F '#{window_zoomed_flag}' 'display \"".msg."\"'"
            \ ." '".args."'"
      let s:tried_moving_out_of_zoomed_tmux = [a:direction, localtime()]
      silent call s:TmuxCommand(args)
    else
      let s:tried_moving_out_of_zoomed_tmux = ['', 0]
      silent call s:TmuxCommand(args)
    endif
```

Still, I think the previous can be improved upon, making it shorter and clearer avoiding state.

oblitum:
The `if` branches between the code that will not switch panes and the code that will. I'm not sure this intention in the code can be made clearer just by writing specific code for each case instead of overwriting variables like `args`. Anyway, it's just a sketch of a fix, maybe it's good enough.

oblitum:
It's a great conf. My approach now is to use `0.2`. Switch on fast double click only.

blueyed:
@oblitum 
I've incorporated your patch and improved it a bit further in general.. this is from a few days ago, but seemed to cause no issues since then (although I've not really paid attention to its behavior).

oblitum:
@blueyed looks great, just tested.

oblitum:
@blueyed heh, I'm not sure whether it's easy to implement but ideally, the message shouldn't be displayed on conscious double click to zoom out on purpose. Maybe showing it up solely after timing out or when detecting a second try happens after it timed out once?

I know maybe it's getting too complex and consider it not a must, just a nasty detail.

blueyed:
@oblitum 
> the message shouldn't be displayed on conscious double click to zoom out on purpose

Then we cannot really display the "Trigger it again in %.1f seconds to zoom out." message to help the user.

I wanted to look into doing a empty `display` to clear it, but at least for me it seems to get cleared when moving out.

I've added a fixup commit though anyway.. ;)

@christoomey 
What's your opinion?

oblitum:
@blueyed it does clear out when moving out in both commits, so the last may have no effect. There was a situation though where the message was left behind a bit longer but I don't recall how to reproduce.

Regardless of that, as I've explained, the ideal non-noisy behaviour would be to wait (cursorhold or timing out) to show the warning and/or when hitting a second time above the timeout threshold, instead of showing it right away always.

As I said, it's liveable with the warning and that would be perfectionism.

blueyed:
> the ideal non-noisy behaviour would be to wait (cursorhold or timing out) to show the warning and/or when hitting a second time above the timeout threshold, instead of showing it right away always.

I agree, but it's not trivial and then the message would need to change obviously.
We can always add this later.

Currently it's still questionable if this gets accepted at all.

oblitum:
I really wish this version to be accepted, this is like the 3rd or so request and I think this implementation is simple enough at the same time it's flexible.

In your last commit did you add anything useful besides trying to tweak the message? I would revert it if makes the code more complex and doesn't provide an effect.

oblitum:
I'd suggest to update the initial description comment to follow the current behaviour.

blueyed:
Without the last commit `msg` would be assigned initially, but not used.
I'm not sure, but it looked at least like some bug.

oblitum:
Ah, ok then.

christoomey:
@blueyed, @oblitum first sorry for my delayed response / radio silence on this.

My initial reaction to this is similar to the other discussions around this behavior.

1. This update introduces a significant amount of new code and complexity (relative the existing total). With all the fun tricks we have to do just to get this thing to work at all, I tend to be very cautious about adopting more.
2. I continue to believe that working within a zoomed tmux pane is an anti-pattern, and as such I don't think we should be enabling it in this plugin.

That said, I've seen this enough times that I wonder if I might be missing something. Is there some user case I'm not seeing that warrants this added functionality? Again, based on the above my initial reaction I'm inclined to close this, but I wanted to first see if there is anything I'm missing. Can you share your thinking

oblitum:
@christoomey, regarding 1, I think the snippet of code added, which has been kept quite self contained, is well worth the functionality it provides.

Regarding 2, isn't that just an opinion? You may consider it's an anti-pattern in your setup but many of us, me included, don't think so.

Maybe showing my programming conditions and patterns you may reconsider. I have a MacBook 13". That's the sole screen I use to program, and it's not big, tmux saves my life, it's like I had several displays in one.

I follow a somewhat consistent pattern, I reserve panes for specific task, e.g., "creating a blog post with [hugo](http://gohugo.io/)":

screenshot from 2016-07-15 22-40-51

Inside pane 2 (which is task specific) I follow another pattern, I split it with a main editing window at the top with Vim open, and other auxiliary windows bellow related to the task but which I access much less frequently, but sometimes I need to. As you can see, I'm using the window bellow to run the local hugo's static web site server.

I follow this pattern a lot. When I'm editing python sometimes I split the window bellow with a [bpython](http://www.bpython-interpreter.org/) REPL and other with the shell open while still having the main editing window at the top.

The main editing window is the one I generally am more interested, and I really need it zoomed most of the time, because my screen is not big. If it was big, I'd probably not care about zooming stuff in and out, I would just let it like that without zooming, but that's how I like to work. Sometimes it's also nice to have the ability of zooming some of the auxiliary windows to check for some long output etc.

I hope to have enlighten the zoom feature a bit.

oblitum:
I've reversed the pane/window terminology above and fixed it.

oblitum:
Besides that, I'm really enjoying the default option implemented here, I don't zoom out by accident, as well as it's easy to zoom out on purpose by double clicking. Originally on #56 I thought about whether it was not a bug to zoom out on navigation, and talked about disabling it. Even thought fully disabling zooming out on navigation was added has an option, some still may prefer that, the default option of zooming on double click is best of both worlds for me.

blueyed:
I have also a 12.5" screen only most of the time.
My use case is more to temporarily zoom a pane: e.g. when going from editing a single file in Vim to using splits etc. in a pane that was using a corner of the screen also before.
And then it is quite annoying to get from full screen/zoomed to a 4th of the screen accidentally (and Vim doing the VimResized dance).
For my main Vim sessions I am using a separate window most of the time already. But this might be split for logs or test output, too - and then I might zoom one of them temporarily.

blueyed:
btw: to zoom windows inside of Vim I occasionally use https://github.com/szw/vim-maximizer (mapped to `<c-w>_`).

oblitum:
I've been using [Dr. Chip's ZoomWin](https://github.com/drn/zoomwin-vim) for that, will check that one.

blueyed:
@oblitum 
I've used that one before, but it was problematic and too

blueyed:
@christoomey 
What do you say to our replies?

dojoteef:
@blueyed, @oblitum I just got caught up on this thread. My PR #151 could be another solution to this issue given a few tweaks I mention in that thread. It does not support the notion of a repeated action as an override like this PR does so if that's something that is desired I am not sure it would be possible with my approach.

@christoomey In regards to zooming I think the single biggest reason people zoom is related to screen real estate just as @oblitum mentioned. On a 13" MacBook it sounds like it's an issue, but it's even more of an issue when you have to work directly on a terminal (without having a virtual terminal in a GUI). While that is not a use case you might have run into it's something that tmux supports well and would be nice to improve in this plugin. You can obviously see from the frequency of that request that it is something others value highly.

blueyed:
I agree in general that we should move the plugin into the direction of fixing the zoom issue(s).

vavaka:
I have added yet another PR #156 that from my point of view does the same thing but in a more straightforward way.

oblitum:
bump

blueyed:
Yeah, I feel like we should move this plugin to https://github.com/vimjas, and get this fixed/addressed after all.

@christoomey does not seem to like it after all (which is fine), but creating a (community) fork is still bad.
Maybe @christoomey agrees with moving it to an org after all?

oblitum:
Vimjas!? Will put that on my resume.

christoomey:
Hey @blueyed @oblitum, first I want to apologize for my absentee nature on this issue. I've not been living up to my role as maintainer, but I intend to return to more regular maintenance and responsiveness.

With regard to a fork, I would very much like to avoid that. I still plan to maintain this repo and am not interested in moving it.

With regard to this specific issue, I'm going to take a more detailed look over the weekend and determine where we should go. Particular to this repo, my primary goal is stability and making it "just work" as much as possible, thus I'm setting a very high bar on what functionality we adopt as all features have a cost in terms of code complexity, ease of debugging, etc. So I'll ask for a few days to review this and make a determination one way or another, and please feel free to bug me if you've not seen a response by mid next week.

blueyed:
@christoomey 
Understood.
Let me know if it should be rebased etc, but it might not be necessary for testing it, and I would rather not do it until you like it in general.

vavaka:
@christoomey don't miss #156 as I think that is less complex and production ready as it does not require refactoring and contains documentation.

christoomey:
I've just merged https://github.com/christoomey/vim-tmux-navigator/pull/165 which provides similar functionality, so I'm going to close this now in favor of that version.

PR #134, opened 2016-06-17
Improve performance with tmate detection

blueyed:
Remove `s:TmuxOrTmateExecutable`: it is rather expensive to call a
system command and not necessary in this case, because Vim can read and
match on the environment variable directly.

To further optimize this, the `match()` could only be done once, but
this is a huge improvement already.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -11,19 +11,6 @@ if !exists("g:tmux_navigator_save_on_switch")
   let g:tmux_navigator_save_on_switch = 0
 endif
 
-function! s:TmuxOrTmateExecutable()
-  if s:StrippedSystemCall("[[ $TMUX == *'tmate'* ]] && echo 'tmate'") == 'tmate'
-    return "tmate"
-  else
-    return "tmux"
-  endif
-endfunction
-
-function! s:StrippedSystemCall(system_cmd)
-  let raw_result = system(a:system_cmd)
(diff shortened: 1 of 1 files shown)


blueyed:
@christoomey you've added it initially in 9dcea0c.  Do I miss something why spawning a shell would be required?!

blueyed:
@christoomey 
Ping.
If you don't object, I'm merging this soonish.

christoomey:
@blueyed sorry for the delay (again!) here. Big 👍 on the speedup, but I'd prefer to keep the extracted function for clarity.

Rather than introduce additional back and forth, I've created a branch with your logic extracted to the named function (https://github.com/christoomey/vim-tmux-navigator/tree/cjt-better-detection-of-tmate). Assuming you have no issues with it, can you `reset` this PR to point at that commit and merge it? I did a quick profile comparing `master`, your version, and mine, and both your original fix and my update are 2x speedup.

Also, unless you have an objection, it would be great if you could work directly against the main repo, rather than opening PRs via your fork. Would make this sort of exchange a bit more straightforward as I could have just pushed a commit to this branch.

(also, thanks again for continued efforts on this project and dealing with my delayed responses :)

blueyed:
> Rather than introduce additional back and forth, I've created a branch

Thanks, but you could have posted a patch also here.  I've merged your branch into master manually now.  No need to update the PR IMHO.
I will push branches to the repo in the future.

PR #138, opened 2016-07-07
Document clear screen without prefix key in single-pane tmux windows, and add example configs to vim-tmux-navigator.tmux

smkent:
For convenience, tmux can be configured to send the clear screen command using `<Ctrl-l>` without requiring the tmux prefix key in tmux windows with only a single pane (i.e., full-terminal tmux windows). This updates `README.md` and the tmux plugin configuration with the necessary binding. The `<Prefix><Ctrl-l>` binding is still required to send the clear screen command in tmux windows with multiple panes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,16 @@ bind C-l send-keys 'C-l'
 
 With this enabled you can use `<prefix> C-l` to clear the screen.
 
+The default `<Ctrl-l>` tmux binding can be further configured to send the clear
+screen command without pressing the tmux prefix key if the current window only
+has a single pane. To enable this behavior, replace the `bind-key -n C-l` line
+in your `~/.tmux.conf` with the following:
+
+``` tmux
+bind-key -n C-l if-shell "$is_vim" "send-keys C-l" \
+    "if-shell \"[ '#{window_panes}' -gt 1 ]\" 'select-pane -R' 'send-keys C-l'"
+```
+
 Thanks to [Brian Hogan][] for the tip on how to re-map the clear screen binding.
 
 #### Nesting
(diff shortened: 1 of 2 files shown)


christoomey:
Hello @smkent, thanks for sharing this. While I certainly appreciate the effort to package this up, my preference has been to avoid functionality like this. Specifically, another user opened [this pr](https://github.com/christoomey/vim-tmux-navigator/pull/126) offering similar functionality. I wrote up my thoughts there, but for completeness, I'll copy them in here:

> While this is a novel solution, I'm very wary of adding any additional complexity to the setup. In addition, I think overloading C-l to do clear screen in addition to the already complex navigation is more than I'm comfortable bringing into the plugin. For my use case I've found the consistency of the prefix C-l mapping to work great and now require any thought.

Please feel free to comment or reopen if you think I've missed something.

README.md
```diff
@@ -134,6 +134,16 @@ bind C-l send-keys 'C-l'
...
 With this enabled you can use `<prefix> C-l` to clear the screen.
 
+The default `<Ctrl-l>` tmux binding can be further configured to send the clear
+screen command without pressing the tmux prefix key if the current window only
+has a single pane. To enable this behavior, replace the `bind-key -n C-l` line
+in your `~/.tmux.conf` with the following:
+
+``` tmux
+bind-key -n C-l if-shell "$is_vim" "send-keys C-l" \
+    "if-shell \"[ '#{window_panes}' -gt 1 ]\" 'select-pane -R' 'send-keys C-l'"
```
blueyed: FWIW, this should also take into account '#[window_zoomed_flag}', and you might want to use `if -F` to not spawn a shell.

blueyed: Ah, in https://github.com/christoomey/vim-tmux-navigator/pull/126/files#r62753664 I've came up with this:

> bind -n C-l if "[ '#{pane_width}' = '#{window_width}' ]" "send-keys C-l" "select-pane -R"

blueyed:
FWIW, I would appreciate a paragraph in the README about it.

smkent:
@blueyed, I really like your solution in https://github.com/christoomey/vim-tmux-navigator/pull/126/files#r62753664. I put it in my own `.tmux.conf`.

PR #139, opened 2016-07-19
Implement universal OS window navigation

thalesmello:
Hey, @christoomey, maybe this pull request might be of interest to the project.

I've always wanted to have `Ctrl + HJKL` to work universally across all windows in
the operating system. So, based on this [blogpost](http://silly-bytes.blogspot.com.br/2016/06/seamlessly-vim-tmux-windowmanager_24.html),
I was able to create a feature branch that implements just that.

As a matter of fact, it already solves my problem. But, since it may of of interest to other users
of this plugin, I've created this pull request.

I've implemented only OSX support, but I think it should be relatively easy to
accomplish the same thing for other GNU/Linux desktop environments (actually,
it should be incredibly simple to adapt the Ratpoison script for which this branch
is based on).

Notice there might be still some rough edges, as I didn't worry about handling
errors in case the user hasn't configured the OS navigation part, so if there is
still interest in this pull request, it might be necessary to polish before merging.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,49 @@ run '~/.tmux/plugins/tpm/tpm'
 Thanks to Christopher Sexton who provided the updated tmux configuration in
 [this blog post][].
 
+#### OS Window Navigator
+
+In case you wish to enable navigation across windows in your operating system,
+there is initial support for Mac OS X with the aid of
+[Hammerspoon](http://hammerspoon.org/). In order to do so, put the following
+code snippet in your `~/.hammerspoon/init.lua`:
+
+``` lua
+require("hs.ipc")
+
+if not hs.ipc.cliStatus(null, true) then
+  hs.ipc.cliInstall()
+end
+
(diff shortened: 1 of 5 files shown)


pedrofranceschi:
:+1

thalesmello:
@christoomey Is there interest in this pull request?

christoomey:
Hey @thalesmello, thanks for sharing this. While I'm certainly impressed by the functionality, my philosophy on this plugin has been to keep things as simple and focused as possible.

With that in mind I'm going to close this, but please feel free to comment or reopen if you think I've missed something. In addition, if you documented this somewhere else I'd be happy add a link to the README.

thalesmello:
I actually agree with you on this one.

Eventually I will want to add Gvim support, which definitely doesn't belong to this project, as it's not Tmux related. Therefore, maybe it makes more sense to start a new project.

Anyway, thank you for your feedback.

PR #143, opened 2016-08-09
Consume error thrown when autosaving file without name

### LINKED ISSUE #142: Autosave error when file not named
If autosave is turned on with `let g:tmux_navigator_save_on_switch = 1` and a vim instance in opened, ie running just `vim` and not `vim <path-to-file>` then navigating away from the tmux pane containing this unnamed (for lack of a better word) vim instance results in the following error:

```
Error detected while processing function <SNR58_TmuxWinCmd..<SNR>58_TmuxAwareNavigate:
line 11:
E32: No file name
```

Here's a [screenshot](https://cloud.githubusercontent.com/assets/1986950/17531318/2f1b321c-5e30-11e6-9e23-7300de4c0dcc.png) of the error.

EDIT: fix formatting

paymog:
Summary: Fixes #142. An issue existed where opening vim without a file
path and then navigating away (with autosave turned on) resulted in an
error being thrown. Now, this error is caught and consumed and not
propagated to the user.

Test Plan:
1) Started up a vim instance without a file path.
2) Created a second tmux pane.
3) navigated to tmux pane with vim
4) entered insert mode and wrote some gibberish
5) navigated away from the tmux pane containing the vim instance
6) no error message was shown to me

Please review the hell out of this, commit message included. It's been a while since I contributed to a project on github and I've literally never written vimscript before.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -68,7 +68,10 @@ function! s:TmuxAwareNavigate(direction)
   " b) we tried switching windows in vim but it didn't have effect.
   if tmux_last_pane || nr == winnr()
     if g:tmux_navigator_save_on_switch
-      update
+      try
+        update
+      catch /^Vim\%((\a\+)\)\=:E32/
+      endtry
     endif
     let args = 'select-pane -' . tr(a:direction, 'phjkl', 'lLDUR')
     silent call s:TmuxCommand(args)


christoomey:
This looks great. Thanks for reporting @paymog. Code looks good (Vim doesn't give you much room to do anything else), and the commit message passes my main criteria of capturing the context / "why" of the code, so I think this looks great.

I did notice that the commit is currently shown as "unknown author", but you can fix this any time by reconnecting the associated email address with your GitHub account.

Thanks for the fix!

blueyed:
Thanks a lot, @paymog!

paymog:
Hey @christoomey, I went ahead and added my work email to my github profile. Has the "unkown author" issue gone away? I may be looking at the wrong screen but it seems that my commit is now authored by me: 
image

PR #144, opened 2016-08-10
Add new option to auto save all buffers on navigation

paymog:
Summary: Added a new option which, when set, causes all open and changed
buffers to be written when navigating away from a vim session. If auto
save is turned on and this option is turned off, only the active buffer
gets saved.

Test Plan:
1) Open up two tmux panes, one running vim and the other not
2) In the vim pane open several files and modify each of them without
explicitly writing any of them
3) navigate away from the tmux split containing vim
4) navigate back to the tmux split containing vim
5) write all files with ":wa"
6) see that nothing got written suggesting that every buffer was written
in step 3

1) enable new option
2) open two tmux splits
3) in one of the splits open vim without a file name. (just `vim` not
`vim <some-file-path>`)
4) enter text into the vim instance
5) navigate away from tmux pnae containing unnamed vim
6) see that no error is shown to user

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,9 +111,19 @@ Left would be created with `nnoremap <silent> <c-h> :TmuxNavigateLeft<cr>`.
 
 ##### Autosave on leave
 
-    let g:tmux_navigator_save_on_switch = 1
+You can configure the plugin to write the current buffer, or all buffers, when navigating from Vim to tmux. This functionality is exposed via the `g:tmux_navigator_save_on_switch` variable, which can have either of the following values:
 
-This will execute the update command on leaving vim to a tmux pane. Default is Zero
+Value  | Behavior
+------ | ------
+1      | `:update` (write the current buffer, but only if changed)
+2      | `:wall` (write all buffers)
+
+To enable this, add the following (with the desired value) to your ~/.vimrc:
+
+```vim
+" Write all buffers before navigating from Vim to tmux pane
+let g:tmux_navigator_save_on_switch = 2
+```
 
 #### Tmux
 
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -67,10 +67,15 @@ function! s:TmuxAwareNavigate(direction)
   " a) we're toggling between the last tmux pane;
   " b) we tried switching windows in vim but it didn't have effect.
   if tmux_last_pane || nr == winnr()
-    if g:tmux_navigator_save_on_switch
+    if g:tmux_navigator_save_on_switch == 1
       try
-        update
-      catch /^Vim\%((\a\+)\)\=:E32/
+        update " save the active buffer. See :help update
+      catch /^Vim\%((\a\+)\)\=:E32/ " catches the no file name error 
+      endtry
+    elseif g:tmux_navigator_save_on_switch == 2
+      try
+        wall " save all the buffers. See :help wall
+      catch /^Vim\%((\a\+)\)\=:E141/ " catches the no file name error 
       endtry
     endif
     let args = 'select-pane -t ' . $TMUX_PANE . ' -' . tr(a:direction, 'phjkl', 'lLDUR')


README.md
```diff
@@ -113,7 +113,11 @@ Left would be created with `nnoremap <silent> <c-h> :TmuxNavigateLeft<cr>`.
 
     let g:tmux_navigator_save_on_switch = 1
 
-This will execute the update command on leaving vim to a tmux pane. Default is Zero
+This will execute the `update` command on leaving vim to a tmux pane. Default is Zero
+
+    let g:tmux_navigator_save_all_buffers_on_switch = 1
```
blueyed: What about using `2` with the existing `g:tmux_navigator_save_on_switch` setting to indicate this behavior?

paymog: Yeah, that seems like a pretty reasonable idea. If you think that's better I'm more than happy to implement it.

paymog: Should we abandon this feature?

paymog:
I'm not sure what standard github etiquette is so I'll ping this PR one last time. If there's no traction by the end of the week I'll just abandon it.

blueyed:
@paymog 
Please be patient.
@christoomey is slow with reviewing stuff here.

As for me, I would squash it into a single option already, and then it can just wait here for more feedback.

README.md
christoomey: I like @blueyed's idea of opting into the existing variable. What do you think of the following for the help text:

> You can configure the plugin to write the current buffer, or all buffers, when navigating from Vim to tmux. This functionality is exposed via the `g:tmux_navigator_save_on_switch` variable, which can have either of the following values:

> | Value | Behavior |
> | ----- | --------- |
> | 1 | `:update` (write the current buffer, but only if changed) |
> | 2 | `:wall` (write all buffers) |
>
> To configure, add the following (with the desired value) to your `~/.vimrc`:
> 
> ``` vim
> " Write all buffers before navigating from Vim to tmux pane
> let g:tmux_navigator_save_on_switch = 2
>  ```

christoomey:
Hey @paymog, sorry for the delays here. As @blueyed fairly pointed out, I can be a bit slow to respond :) I can't speak for other folks managing projects, but I'm personally fine if you ping after not hearing anything for > a week. 

That said, I think this is great! I like @blueyed's idea re: using the existing variable and I've added a note about possible wording for the REAMDE section about it.

Steps to get this merged:

1) Update to use the existing variable
2) Squash down to a single commit (and if you're ok with it I'd remove the "test plan" from the commit message. Makes sense for PR discussion, but doesn't need to live on in the log. Want git log to focus on the "why" of a change)
3) Ping me one more time (I'll keep an eye out for it) :)

Thanks again, @paymog and @blueyed

paymog:
Sorry for bothering you guys so much! Didn't mean to be a pain in the ass. 

The updated PR should be out now. I promise I won't pester you guys anymore now that I know it's a slow process :)

blueyed:
> Didn't mean to be a pain in the ass. 

Don't worry - you are not.. :)

Let's see if @christoomey holds on to his promises.. ;)

README.md
```diff
@@ -111,9 +111,19 @@ Left would be created with `nnoremap <silent> <c-h> :TmuxNavigateLeft<cr>`.
...
 ##### Autosave on leave
 
-    let g:tmux_navigator_save_on_switch = 1
+You can configure the plugin to write the current buffer, or all buffers, when navigating from Vim to tmux. This functionality is exposed via the `g:tmux_navigator_save_on_switch` variable, which can have either of the following values:
 
-This will execute the update command on leaving vim to a tmux pane. Default is Zero
+Value  | Behavior
+------ | ------
+1 | `:update` (write the current buffer, but only if changed)
+2 | `:wall` (write all buffers)
```
blueyed: Can this be aligned to be better readable also in plain text (without being rendered)?

README.md
```diff
@@ -111,9 +111,19 @@ Left would be created with `nnoremap <silent> <c-h> :TmuxNavigateLeft<cr>`.
...
+
+To configure, add the following (with the desired value) to your ~/.vimrc:
```
blueyed: s/To configure,/To enable this/

plugin/tmux_navigator.vim
```diff
@@ -67,12 +67,16 @@ function! s:TmuxAwareNavigate(direction)
...
-    endif
+    try
+      if g:tmux_navigator_save_on_switch == 1
+          update " save the active buffer. See :help update
+      elseif g:tmux_navigator_save_on_switch == 2
+          wall "  save all the buffers. See :help wall
+      endif
+    catch /^Vim\%((\a\+)\)\=:E32/ " catches the no file name error thrown by update
+    catch /^Vim\%((\a\+)\)\=:E141/ " catches the no file name error thrown by wall
+    endtry
```
blueyed: You might want to put the command in their own `try` blocks (and only catch the relevant error).

blueyed:
Thanks, LGTM - given some minor doc fix and other nitpicking.. ;)

Feel free to push this as a fixup commit (http://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html), and we can then review it better (only the changes), and squash it from GitHub's UI.

README.md
```diff
@@ -111,9 +111,19 @@ Left would be created with `nnoremap <silent> <c-h> :TmuxNavigateLeft<cr>`.
...
+
+```
```
christoomey: Can you add `vim` after the backticks so syntax highlighting kicks in?

christoomey:
@paymog No worries about bugging, you are not. I'm just a bit slower than we'd all probably like :)

All of @blueyed's comments sounds good to me, plus I added one more. Once we get those in we can get this merged. Thanks for continued effort.

plugin/tmux_navigator.vim
paymog: Great idea!

blueyed:
Awesome! :+1:

paymog:
@christoomey ping!

christoomey:
@paymog merged! Thanks for continued efforts on this.

paymog:
Thanks! Feels good to contribute back!

PR #145, opened 2016-08-12
Fix switching panes when multiple sessions are attached to the same window(s).

jamesoff:
I run several tmux sessions all pointed at the same source session (one in a windowed terminal, one in fullscreen, and sometimes one via ssh from another machine). This patch uses the `TMUX_PANE` environment variable in the `select-window` command to help tmux understand which window you want to change the selection in. Otherwise, the command only works from one of the sessions.

(I did look at fetching it with `display-message -p '{#D}'` but oddly that seemed to return nothing sometimes.)

My first vimscript hacking, hope I did it right!

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -73,7 +73,7 @@ function! s:TmuxAwareNavigate(direction)
       catch /^Vim\%((\a\+)\)\=:E32/
       endtry
     endif
-    let args = 'select-pane -' . tr(a:direction, 'phjkl', 'lLDUR')
+    let args = 'select-pane -t ' . $TMUX_PANE . ' -' . tr(a:direction, 'phjkl', 'lLDUR')
     silent call s:TmuxCommand(args)
     if s:NeedsVitalityRedraw()
       redraw!


plugin/tmux_navigator.vim
```diff
@@ -73,7 +77,7 @@ function! s:TmuxAwareNavigate(direction)
       catch /^Vim\%((\a\+)\)\=:E32/
       endtry
     endif
-    let args = 'select-pane -' . tr(a:direction, 'phjkl', 'lLDUR')
+    let args = 'select-pane -t ' . s:TmuxPaneId() . ' -' . tr(a:direction, 'phjkl', 'lLDUR')
```
blueyed: I would just inline `$TMUX_PANE` here.

jamesoff:
Updated to do that :)

blueyed:
:+1:

blueyed:
Let's wait for feedback from @christoomey / others.

christoomey:
@jamesoff Thanks for the update. I tested locally and it worked fine for me. Merged into master.

@blueyed thanks as always for your speedy reviews!

Issue #148, opened 2016-08-31
Can't move from vim to tmux with fish as shell


scolby33:
I am having a similar issue as #141, but in my case `:TmuxPaneCurrentCommand` does output `vim`. Otherwise, the plugin and configuration work perfectly: movement between vim splits, movement between tmux panes, and movement from tmux panes to vim.

I know it is not the key bindings; running `:TmuxNavigate{Up|Left|Down|Right}` manually moves between vim splits successfully but will **not** move out of vim into tmux.

OS X 10.11.6
Vim 7.4
tmux 2.2
iTerm2 3.0.7

christoomey:
@scolby33 sorry to hear about the trouble. Based on the fact that you can move within splits, I don't believe the Vim detection is the problem. Instead, it sounds like the Vim plugin code is not properly detecting when you're at the wall. Can you describe what happens when you attempt to navigate from Vim to tmux? Is it nothing, or do you move to a different Vim split?

Also, would you mind running with just the plugin active (no vimrc) using the following command:

``` vim
vim -U NONE -c 'source ~/.vim/bundle/tmux-navigator/plugin/tmux_navigator.vim'
```

scolby33:
If I am already at the "edge" of vim and attempt to navigate out to tmux, nothing happens.

Running with just the plugin active did not solve the issue.

However, I **am** using `fish` as my shell. Adding the following snippet to my .vimrc fixed the issue:
```vim
if &shell =~# 'fish$'
    set shell=sh
endif
```

This is a bit surprising because (in only a cursory look, admittedly), I don't see why the plugin's attempts to call out to the shell would be hindered by `fish` vs. `bash` or `sh`...the tmux commands are the same for all.

christoomey:
Depending on the version of the plugin you're using, we might have introduced an env variable into the `system` call that could be the issue. Can you try running `:echo $TMUX_PANE` with fish as your shell to confirm?

scolby33:
With fish as the shell, this returns `%1`. The same result occurs when forcing the shell to be sh as well.

I am using the most recent plugin version installed by Vundle.

christoomey:
Can you try running the following?

1. Tmux window with Vim on the left, tmux pane on the right
2. In Vim, run `:echo system('tmux select-pane -t '. $TMUX_PANE .' -R')` and capture anything that is echoed
3. Check the output (again in Vim) of `:echo v:shell_error`.

scolby33:
[...]

PR #150, opened 2016-09-02
Escape the % in the tmux select-pane -t command to prevent process substitution by fish

scolby33:
Re #148

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -78,7 +78,7 @@ function! s:TmuxAwareNavigate(direction)
       catch /^Vim\%((\a\+)\)\=:E141/ " catches the no file name error 
       endtry
     endif
-    let args = 'select-pane -t ' . $TMUX_PANE . ' -' . tr(a:direction, 'phjkl', 'lLDUR')
+    let args = 'select-pane -t \' . $TMUX_PANE . ' -' . tr(a:direction, 'phjkl', 'lLDUR')
     silent call s:TmuxCommand(args)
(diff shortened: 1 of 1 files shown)


plugin/tmux_navigator.vim
```diff
@@ -78,7 +78,7 @@ function! s:TmuxAwareNavigate(direction)
       catch /^Vim\%((\a\+)\)\=:E141/ " catches the no file name error 
       endtry
     endif
-    let args = 'select-pane -t ' . $TMUX_PANE . ' -' . tr(a:direction, 'phjkl', 'lLDUR')
+    let args = 'select-pane -t \' . $TMUX_PANE . ' -' . tr(a:direction, 'phjkl', 'lLDUR')
```
blueyed: What about `let args = 'select-pane -t ' . shellescape($TMUX_PANE) . ' -' . tr(a:direction, 'phjkl', 'lLDUR')` instead?

blueyed: i.e. use `shellescape()` instead of the backslash.

scolby33: Good call. It seems that enclosing in quotes works as well as escaping with `\`.

scolby33:
Let me know if you want me to squash commits before merging.

blueyed:
:+1: 

I prefer it squashes, yes.

But we can do so from the GitHub UI.  Feel free to force-push what you like, of course.

Please ping me - giving also time for @christoomey to chime in, but I think this is good to fly.

blueyed:
:fireworks:

PR #151, opened 2016-09-18
Added nested navigation support!

dojoteef:
* Use global option @navigation-bounded to force navigation to not wrap
during motion commands (default 'off').

* Use global option @navigation-nested to allow navigating nested tmux
sessions. This option implies @navigation-bounded as it is required for
proper nested navigation behavior (default 'off').

* Use global option @navigation-delay to specify a larger or smaller
delay between issuing the movement command and executing it over ssh.
Larger values might be needed if high latency ssh sessions are in use.
(default '0.2').

This has been tested both on locally nested tmux sessions (using `TMUX= tmux`
to get around nested sessions warnings) and on nested tmux/ssh
sessions. It seems to work fine on two levels of nesting, but I have
noticed quirks at higher nesting levels. I have not put in the time and
effort to fix those as I do not currently have a usecase.

Thanks so much for the work you put in for this great plugin. I do not have too
much time to help usher this pull request through. In fact I still have a mini
feature I made in my .vimrc for [neomake](https://github.com/neomake/neomake) that I have not had the time to package
up into a pull request. That said I will try my best.

Hopefully with this feature you will get to remove this from your README:
> If you like to nest your tmux sessions, this plugin is not going to work properly.
> It probably never will...

😉

### DIFF
--- a/scripts/helpers/direction_utils.sh
+++ b/scripts/helpers/direction_utils.sh
@@ -1,0 +1,115 @@
+function window_height() {
+  tmux display-message -p "#{window_height}"
+}
+
+function window_width() {
+  tmux display-message -p "#{window_width}"
+}
+
+function pane_top() {
+  tmux display-message -p "#{pane_top}"
+}
+
+function pane_bottom() {
+  tmux display-message -p "#{pane_bottom}"
+}
+
+function pane_left() {
+  tmux display-message -p "#{pane_left}"
+}
+
+function pane_right() {
+  tmux display-message -p "#{pane_right}"
+}
+
+function pane_index() {
+  tmux display-message -p "#{pane_index}"
+}
+
+function is_bounded() {
+  tmux show-options -g @navigation-bounded &> /dev/null || return 1
+}
+
+function can_select_pane() {
+  if is_bounded || is_nested; then
+    case $1 in
+      L)
+        [[ $(pane_left) -gt 0 ]]
+        ;;
+      R)
+        [[ $(pane_right) -lt $(($(window_width) - 1)) ]]
+        ;;
+      U)
+        [[ $(pane_top) -gt 0 ]]
+        ;;
+      D)
+        [[ $(pane_bottom) -lt $(($(window_height) - 1)) ]]
+        ;;
+    esac
+  fi
+}
+
+function select_pane() {
+  case $1 in
+    L)
+      is_nested && select_pane_left || tmux select-pane -L
+      ;;
+    R)
+      is_nested && select_pane_right || tmux select-pane -R
+      ;;
+    U)
+      is_nested && select_pane_up || tmux select-pane -U
+      ;;
+    D)
+      is_nested && select_pane_down || tmux select-pane -D
+      ;;
+  esac
+}
+
+function select_pane_dir() {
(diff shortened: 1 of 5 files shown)


blueyed:
Woah! Thanks!

Have you seen https://github.com/christoomey/vim-tmux-navigator/pull/133 - it seems to go into the same direction.

dojoteef:
@blueyed I just took a look. It seems to be solving a different problem (zooming) and doing so by requiring a repeated action (similar to pressing the prefix key twice to send the key to the current session).

That said you reference #56, #65, #104 in your zooming PR (#133). I had not seen them before, but I believe my code would address the concerns listed in those issues. If you are in a zoomed tmux pane and have the `@navigation-bounded` flag set then it will disallow moving. If that flag is set the code will only allow moving if there is a visible pane that can be moved into in the desired direction, since no other panes are visible due to the zoom you would stay inside the current pane. That functionality is required to get nested navigation to work correctly, but I figured it could be broken out into its own setting in case someone was not interested in nested navigation though still wanted the bounding effect.

dojoteef:
So there are a couple of issues in this PR that probably need to be addressed before it gets merged.

1.  I never use C-/ to go to the last pane so I didn't realize I had broken that functionality. That one seems more difficult to support across nested sessions so I would likely just restore the previous functionality for it.
2. If the vim issue is big enough that people wanted it addressed it could be easily solved by having the vim script call the bash scripts I wrote to do the navigation rather than calling `tmux` directly. That way the behavior in vim would be consistent with the behavior outside of vim without needing any additional flags for people to mess with. Doing this **should** address the issues around navigation within a zoomed pane that are listed in my previous comment.

blueyed:
I've just skimmed the PR's diff and your comments, but it looks like it is forking out too much: from all the`tmux` calls and starting to use libbraries in .sh...

I get the feeling that #133 is the way to go regarding simplicity in general, but with additional fixed from here.  What do you think?
(but being that again: I've only skimmed it)

dojoteef:
@blueyed I am sorry, but I am confused by your comment. As I stated before PR #133 is trying to solve a different issue. The title of my PR is nested navigation support, i.e. it allows you to navigation between nested sessions of tmux. It just so happens that my code also incidentally could work to address #133, but that's not it's primary purpose.

As for forking too much/using libraries in .sh, I'm not sure I understand your meaning. I'm not using any libraries and I am simply using bash and the command line interface to tmux, which is no different than what the current bash script does see [vim-tmux-navigation.tmux](https://github.com/christoomey/vim-tmux-navigator/blob/d5dc8d3c80cf898a25b8cf16ab65cf6a5438009f/vim-tmux-navigator.tmux#L5-L9)

The only issue regarding portability would be my use of `sleep` as traditionally `sleep` only accepts whole numbers for the seconds parameter. As far as I know there is no universally portable way to sleep for a certain number of milliseconds, though GNU sleep supports and I believe the default OSX sleep does as well.

That said if this is not a PR that is desired for this plugin that's fine. What I've written is probably as much or more code than already existed in this plugin so I understand the hesitance in accepting it. I just figured it might be useful to others, but I'd be happy to make this available as a new plugin unrelated to vim-tmux-navigator.

PR #152, opened 2016-09-26
do not move to other tmux pane in zoomed mode

vavaka:
In zoomed mode accidental pressing on a navigation shortcut brings me back to unzoomed mode and navigates to another pane. To restore my window state I have to navigate back to previous pane and zoom it. As a user in zoomed mode I want to be focused on a single pane only, I don't care about other panes and their location in unzoomed mode. That's why I want pane navigation shortcuts do not work in zoomed mode.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -57,6 +57,11 @@ function! s:NeedsVitalityRedraw()
   return exists('g:loaded_vitality') && v:version < 704 && !has("patch481")
 endfunction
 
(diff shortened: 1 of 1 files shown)


christoomey:
Hey @vavaka, thanks for the update. There is currently [another PR](https://github.com/christoomey/vim-tmux-navigator/pull/133) in works that adds an optional configuration like what you're suggesting here. I'm going to close this in favor of that one to keep the work focused, so I'd suggest tracking that issue for updates.

PR #156, opened 2016-10-16
Configurable moving out of zoomed tmux pane

vavaka:
Hi

Initially I created PR #152. There I explained why people may need to prevent moving out of zoomed tmux pane. #152 had some shortcomings that are fixed in this PR:
- added option for switching on/off moving out of zoomed tmux pane.
- documentation added.
- code made more clear and more readable.

I know that there is a PR #133 which is doing the same but I think that solution from this PR has some disadvantages:
1. key press timeouts is a complex approach, it would be better to keep things simple and straightforward by just having ability to enable/disable moving out from zoomed tmux pane.
2. code is complex and not readable

From #133 I know that some people think that working in a zoomed tmux pane is anti-pattern. I completely disagree with this statement, I find pane zooming extremely useful on small size screens. From #133 I know that there are people who also find zoom feature useful. I think that it's person's choice to decide if he or she want to use this feature or not. The problem is that pane zooming is a tmux feature and vim-tmux-navigator handles it only in a single way while there are at least 2 ways. Moreover, I find prevent moving out from zoomed pane more natural then allowing to do it. When I'm in a zoomed pane I want to focus on this pane only and I don't want to accidentally leave it. The only reason for zooming out should be explicit call of `zoom out` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,12 @@ Alter each of the five lines of the tmux configuration listed above to use your
 custom mappings. **Note** each line contains two references to the desired
 mapping.
 
+#### Moving out of zoomed Tmux pane
+
+    let g:tmux_navigator_zoom_out_navigation = 1
+
+This will enable moving out of zoomed Tmux pane. Default is `0`.
+
 ### Additional Customization
 
 #### Restoring Clear Screen (C-l)
--- a/doc/tmux-navigator.txt
+++ b/doc/tmux-navigator.txt
@@ -24,6 +24,9 @@ CONFIGURATION                             *tmux-navigator-configuration*
 * Custom Key Bindings
  let g:tmux_navigator_no_mappings = 1
 
+* Activate moving out from zoomed tmux pane
+ let g:tmux_navigator_zoom_out_navigation = 1
+
  nnoremap <silent> {Left-mapping} :TmuxNavigateLeft<cr>
  nnoremap <silent> {Down-Mapping} :TmuxNavigateDown<cr>
  nnoremap <silent> {Up-Mapping} :TmuxNavigateUp<cr>
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -11,6 +11,10 @@ if !exists("g:tmux_navigator_save_on_switch")
   let g:tmux_navigator_save_on_switch = 0
 endif
 
+if !exists("g:tmux_navigator_zoom_out_navigation")
+  let g:tmux_navigator_zoom_out_navigation = 0
+endif
+
 function! s:TmuxOrTmateExecutable()
   return (match($TMUX, 'tmate') != -1 ? 'tmate' : 'tmux')
 endfunction
@@ -38,10 +42,10 @@ function! s:TmuxPaneCurrentCommand()
 endfunction
 command! TmuxPaneCurrentCommand call s:TmuxPaneCurrentCommand()
 
-let s:tmux_is_last_pane = 0
+let s:last_navigated_to = "vim"
 augroup tmux_navigator
   au!
-  autocmd WinEnter * let s:tmux_is_last_pane = 0
+  autocmd WinEnter * let s:last_navigated_to = "vim"
 augroup END
 
 " Like `wincmd` but also change tmux panes instead of vim windows when needed.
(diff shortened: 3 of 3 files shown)


blueyed:
Thanks!
I like #133 better though.. ;)

The code can be simplified, but first we would need to convince (or overrule?) @christoomey  anyway.

oblitum:
Also prefer #133.

vavaka:
@blueyed @oblitum Thank's for your response. I respect your opinion but from my point of view there should be the only one way to leave zoom mode - explicitly zoom out. As a user I don't want window be zoomed out by accidentally pressing window/pane navigation keys (in the case #133 I can do it by pressing navigation key twice).

Moreover, I think that it is more natural to limit pane navigation in zoomed mode because the main idea of zoomed mode is focusing on a single pane, in this mode pane layout makes no sense because there is only a one (zoomed) pane at the moment. How user can take decision what pane he wants navigate to if he does not see panes layout?

That's why I think that zooming out on double key pressing is not natural, complex to understand and complex to implement approach.

blueyed:
> As a user I don't want window be zoomed out by accidentally pressing window/pane navigation keys (in the case #133 I can do it by pressing navigation key twice).

This can be configured.

> How user can take decision what pane he wants navigate to if he does not see panes layout?

From memory?!

> That's why I think that zooming out on double key pressing is not natural, complex to understand and complex to implement approach.

Yes, I agree in that regard.  But it's implemented already, and just needs a decision on the default and then some code cleanup maybe.

vavaka:
>This can be configured.
Can you provide with a sample configuration?

> From memory?!
In this case if I were a user who wants to navigate in zoomed mode between panes, I would want to stay in zoomed mode after navigating to another pane. From my point of view it is more convenient then automatically zooming out on navigation.

oblitum:
Hi @vavaka. You should please take your time to read a competing pull-request that has been standing this long. Until now there are still 2 votes against 1 and the previous PR covers your case and it's clearly (I originally was thinking about a feature just like you're suggesting but considered @blueyed solution an improvement). You're basically discussing non-inclusion of something that won't affect you since you can just configure the functionality as you're already describing.

vavaka:
@oblitum Thank you for reply. You and @blueyed did a great job but as a vim-tmux-navigator user I find your solution complex, inconsistent and not completed. Let me explain it one more time.
1. I find #133 complex because of timeout logic for handling navigation keys twice pressing. I think that it makes complex plugin codebase, configuration and future support.
2. I find #133 inconsistent because it adds "zoom out" side effect when user navigates to another pane in zoomed mode. I think that staying in zoomed mode will be more consistent.
3. I find #133 not completed because of lack of documentation about configuration values you added. Also I think code added in your solution requires refactoring because at the moment it looks not good.

Saying about #156, I think that it is quite straightforward and easy to understand, it has no side effects, it is documented, code is refactored and well structured.

Regarding votes, I think that it is up to @christoomey decide what changes to accept. Votes mean nothing, only common sense makes sense. I will respect any of his decisions.  I don't care about whose changes will be applied. All I want is to make work with this plugin more handy.

oblitum:
@vavaka if you didn't notice, there was no more time investment into #133 (docs, etc) after basic functionality was achieved because opinion/review by @christoomey was being expected first.

vavaka:
@oblitum I don't think that code refactoring and adding documentation to #133 could take a lot of your time. Moreover it would make your solution complete and help project maintainer to see the entire picture before taking decision about your pull request.

oblitum:
@blueyed it's not my pull request. I simply defended that one because its functionality covers your use case as well as my actual one (and others). Actual because initially it was your use case that was over my mind solely, on #56, but #133 made me change my mind for something I found better. The complexity added for it is simply negligible.

oblitum:
@vavaka  it's not my pull request. I simply defended that one because its functionality covers your use case as well as my actual one (and others). Actual because initially it was your use case that was over my mind solely, on #56, but #133 made me change my mind for something I found better. The complexity added for it is simply negligible.

christoomey:
I've just merged https://github.com/christoomey/vim-tmux-navigator/pull/165 which provides similar functionality, so I'm going to close this now in favor of that version.

PR #165, opened 2017-02-01
Add option to disable tmux navigation while zoomed

christoomey:
Users have requested the ability to disable navigation while the Vim
tmux pane is zoomed, to prevent accidentally unzooming the pane when
navigating past the edges.

This change adds an option to prevent using tmux navigation when the Vim
pane is zoomed, and requiring an explicit unzooming of the tmux pane
before the plugin resumes its magic.

-------

Preventing navigation has been requested a number of times in this repo (#56, #104, #65, #151, and most recently #133 and #156).

I've consistently resisted adding support for this as I felt that working within a zoomed pane is not an optimal workflow and would add undue complexity, but based on the quantity of requests for this feature, I realize this is something that a number of people want, and this code is my attempt to provide a solution while adding as little complexity as possible.

I've opted for a simple and direct implementation, largely based on @glentakahashi's solution presented in #104. To sum up the changes:

- `g:tmux_navigator_disable_when_zoomed` is provided as a new option
- The config is off by default so as to not be a breaking change
- When set to `1`, we prevent navigating out of Vim when zoomed
- To return to normal behavior, the user unzooms the pane

Thanks all for the contributions, suggestions, and patience :)

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -11,6 +11,10 @@ if !exists("g:tmux_navigator_save_on_switch")
   let g:tmux_navigator_save_on_switch = 0
 endif
 
+if !exists("g:tmux_navigator_disable_when_zoomed")
+  let g:tmux_navigator_disable_when_zoomed = 0
+endif
+
 function! s:TmuxOrTmateExecutable()
   return (match($TMUX, 'tmate') != -1 ? 'tmate' : 'tmux')
 endfunction
@@ -21,6 +25,10 @@ endfunction
 
 function! s:InTmuxSession()
   return $TMUX != ''
+endfunction
+
+function! s:TmuxVimPaneIsZoomed()
+  return s:TmuxCommand("display-message -p '#{window_zoomed_flag}'") == 1
 endfunction
 
 function! s:TmuxSocket()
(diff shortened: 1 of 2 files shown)


christoomey:
@blueyed This does not have all the functionality of your version, but my driving concern is simplicity so I opted for this. Do you think this covers the use cases sufficiently? @vavaka, @oblitum Will this version work for you?

oblitum:
From your bullets it seems also similar to @vavaka's proposion. Sadly for me it won't work as I'm already used to the niceties provided by #133. Also, #133 is another one to the list that was not referred. Currently it's what I'm doing while I've been waiting it to merge, staying up to date the to the rare updates in master while merging #133. If it's not going to be merged since it covers all cases I guess I'll live with a fork. The changes applied in this one though seem much minor than #156, but achieves the same effect.

christoomey:
@oblitum thanks for the notes. I've updated to correct typo referencing 131 instead of 133 as intended.

In the end I think the functionality in 133 is interesting, but for me as maintainer the complexity outweighs the utility. I'm hopeful that the code included in this PR will suffice for 95% of use cases, but I'm sorry if it's not quite enough for you.

plugin/tmux_navigator.vim
```diff
@@ -57,25 +65,33 @@ function! s:NeedsVitalityRedraw()
   return exists('g:loaded_vitality') && v:version < 704 && !has("patch481")
 endfunction
 
+function! s:ShouldForwardNavigationBackToTmux(tmux_last_pane, at_tab_page_edge)
+  if s:TmuxVimPaneIsZoomed() && g:tmux_navigator_disable_when_zoomed
```
blueyed: The conditions should be swapped.

blueyed:
@oblitum 
Maybe you can put Vimjas on your resume now shortly.. ;)
I will keep using my version, too, and then probably put it there, but keep it under my username for now. Feel free to create a PR there in case you have local changes already.

oblitum:
@blueyed nice :) I don't have any change besides having rebased it with master lately, it just embeds a single line from resolution conflict, if your fork will keep alive I may grab from it.

christoomey:
@blueyed @oblitum thanks for the additional feedback, as well as your patience on this issue. I've merged in this version providing a minimal support for the disable functionality. If you decide that this can work for you then that is great, but I understand if you've become comfortable with the timeout behavior and stick to that.

PR #170, opened 2017-03-16
Add option to issue display-panes command on leaving Vim

wilriker:
This makes it a lot easier to find out where you land coming from Vim if there are multiple splits below/beside of Vim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,15 @@ To disable navigation when zoomed, add the following to your ~/.vimrc:
 let g:tmux_navigator_disable_when_zoomed = 1
 ```
 
+##### Use display-panes Command on leave
+
+You can configure the plugin to also emit the `display-panes` command to tmux when leaving Vim. This will display an overlay helping to find the target (e.g. in case there are multiple splits below/beside Vim).
+
+```vim
+" Display overlay with target pane
+let g:tmux_navigator_display_panes = 1
+```
+
 #### Tmux
 
 Alter each of the five lines of the tmux configuration listed above to use your
--- a/doc/tmux-navigator.txt
+++ b/doc/tmux-navigator.txt
@@ -24,6 +24,9 @@ CONFIGURATION                             *tmux-navigator-configuration*
 * Custom Key Bindings
  let g:tmux_navigator_no_mappings = 1
 
+* Activate display-panes command on switching out of vim
+ let g:tmux_navigator_display_panes = 1
+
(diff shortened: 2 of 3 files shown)


blueyed:
Maybe it would be better to have a generic hook there to handle custom (tmux) commands?

wilriker:
@blueyed This sounds great but introduces a whole new level of complexity in implementation or usage of this feature because either 

1. you put all the responsibility on the user to provide correct commands to prepend/append (including necessary escapings depending on the shell used - see #148) as well as separators (`\;`) after prepending commands and before appending commands or
2. you would accept e.g. a list of commands to prepend/append, join them together correctly with the necessary separator but then you would have the complexity of finding characters that need to be escaped which is more or less bound to fail

Do you have any advice on how to handle this?

Best regards
Manuel

christoomey:
Hey @wilriker, thanks for sharing this. Unfortunately I think this sort of goes against the main goal of the project (providing "seamless" navigation between vim and tmux).

That said, I would hate to leave you hanging without a solution to your use case, so I'm wondering if an alternative approach might work for you. Specifically, it's possible to enable dimming of inactive panes in tmux / vim ([reference blog post](http://www.deanbodenham.com/learn/tmux-pane-colours.html)) which I think could serve a similar purpose, but not require any changes in the tmux navigator functionality.

wilriker:
Hey @christoomey, it's interesting how differently the term "seamless" can be interpreted. In my setup I have the `display-panes` command whenever I switch between `tmux` panes. I got that from when I was testing `byobu`, though I did not stay with it, I took some of the commands and ideas from it and `byobu` makes heavy use of `display-panes`.
It took me a while to figure out what was different when switching from `vim` to `tmux` versus switching between `tmux` panes and _into_ `vim` because there was something not "seamless" happening. ;-)

But thanks for providing this alternative method of dimming inactive panes. I might try that sometime. Until then I will probably stay with my fork of your plugin, if you don't mind. ;-)

Best regards
Manuel

blueyed:
@wilriker 
FYI: there is probably some maintained fork happening anyway at https://github.com/Vimjas/vim-tmux-navigator, mainly for https://github.com/christoomey/vim-tmux-navigator/pull/133.

oblitum:
@blueyed FWIW, #133 is not in that fork.

blueyed:
@oblitum 
Yeah, thanks for letting me know.
I (or anybody else) hasn't done it yet - I am still using a local branch myself currently.

Issue #140, opened 2016-07-27
Custom mappings fail with Tmux, Neovim, Nixos


Meyermagic:
I doubt this is actually issue with your plugin, but I'd appreciate any help debugging the issue.

My window navigation bindings work in vim alone, and tmux alone (and behave identically), but vim nested inside tmux does not respond at all. Calling the TmuxNavigate* commands directly works as expected.

tmux 2.2
Neovim 0.1.4, packaged via https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/editors/neovim/default.nix

~/.tmux.conf
`# Smart pane switching with awareness of Vim splits.
# See: https://github.com/christoomey/vim-tmux-navigator
is_vim="ps -o state= -o comm= -t '#{pane_tty}' | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|(\.?n?vim?x?(-wrapped)?))(diff)?$'"

# Ctrl-arrow navigation
bind-key -n C-Left if-shell "$is_vim" "send-keys C-Left"  "select-pane -L"
bind-key -n C-Down if-shell "$is_vim" "send-keys C-Down"  "select-pane -D"
bind-key -n C-Up if-shell "$is_vim" "send-keys C-Up"  "select-pane -U"
bind-key -n C-Right if-shell "$is_vim" "send-keys C-Right"  "select-pane -R"

# Ctrl-shift-arrow resizing
bind-key -n C-S-Left if-shell "$is_vim" "send-keys C-S-Left"  "resize-pane -L"
bind-key -n C-S-Down if-shell "$is_vim" "send-keys C-S-Down"  "resize-pane -D"
bind-key -n C-S-Up if-shell "$is_vim" "send-keys C-S-Up"  "resize-pane -U"
bind-key -n C-S-Right if-shell "$is_vim" "send-keys C-S-Right"  "resize-pane -R"

# Alt-arrow splitting (create in direction of arrow)
bind-key -n M-Down if-shell "$is_vim" "send-keys M-Down"  "split-window -v"
bind-key -n M-Up if-shell "$is_vim" "send-keys M-Up"  "split-window -v; select-pane -U"
bind-key -n M-Left if-shell "$is_vim" "send-keys M-Left"  "split-window -h; select-pane -L"
bind-key -n M-Right if-shell "$is_vim" "send-keys M-Right"  "split-window -h"
`

[...]

PR #171, opened 2017-04-02
Detect Nix wrapped vim

NickHu:
This plugin doesn't work in Nix/NixOS because of the way that binaries are
wrapped:

```
$ ps -o state= -o comm= -t /dev/pts/9
S zsh
S .nvim-wrapped
S python3
```

This change tweaks the pattern slightly to accommodate that.

### DIFF
--- a/vim-tmux-navigator.tmux
+++ b/vim-tmux-navigator.tmux
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
-    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
+    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|\.?n?vim?x?(-wrapped)?)(diff)?$'"
 tmux bind-key -n C-h if-shell "$is_vim" "send-keys C-h"  "select-pane -L"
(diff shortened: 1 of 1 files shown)


NickHu:
See also https://github.com/christoomey/vim-tmux-navigator/issues/140

NickHu:
cc @christoomey

christoomey:
Hi @NickHu, thanks for sharing this. While I certainly appreciate your work here to make this plugin more portable, I'm very wary of making more changes to the vin detection pattern as it is already one of the more complex and potentially brittle pieces of this plugin, but unlike the Vim plugin, the tmux side of things is much easier to modify on a case by case basis (although unfortunately this will require a non-TPM mamual configuration).

In the end, I think the right thing for me to do would be to document the process of augmenting the pattern so it is more straightforward for edge cases like nixos.

NickHu:
For people finding this now, it seems that in NixOS unstable, the neovim (+ vim?) process is no longer `-wrapped` in `ps`, making this patch redundant.

christoomey:
Good to know. Thanks for following up and sharing that, @NickHu!

BjornMelgaard:
this fix works perfectly in my nixos, why not accept it?

christoomey:
@BjornMelgaard please see my summary in https://github.com/christoomey/vim-tmux-navigator/pull/171#issuecomment-315064855 for logic around keeping this as user configuration, rather than adding code to support it.

PR #175, opened 2017-04-21
[RFC] implement tmux 2.4 compatibility, fixes #159

### LINKED ISSUE #159: tmux 2.4 compatibility
See:

https://github.com/tmux/tmux/commit/76d6d3641f271be1756e41494960d96714e7ee58
https://github.com/tmux/tmux/issues/592

vim-tmux-navigator bindings now fail in tmux 2.4 when panes are in `copy-mode-vi`.

atweiden:
See: [changes from 2.3 to 2.4 20 April 2017](https://github.com/tmux/tmux/blob/master/CHANGES). The bindings <kbd>C-h</kbd>, <kbd>C-j</kbd>, <kbd>C-k</kbd> and <kbd>C-l</kbd> are broken in tmux 2.4 *copy-mode*.

This code is a step up from having your cursor frozen in Vim while in tmux 2.4 copy-mode, but I had to remove the `if-shell` part of the script. Don't know why.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,11 @@ bind-key -n C-j if-shell "$is_vim" "send-keys C-j"  "select-pane -D"
 bind-key -n C-k if-shell "$is_vim" "send-keys C-k"  "select-pane -U"
 bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
 bind-key -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
+bind-key -T copy-mode-vi C-h select-pane -L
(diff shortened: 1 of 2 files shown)


christoomey:
Hi @atweiden, thanks for the update. I'm not on tmux 2.4 yet myself so I can't test the correct behavior, but I've tested and there are no issues with this running on older tmux (specifically 2.3 in my case), and it sounds like a nice fix for tmux 2.4 users.

The removal of the `if-shell` seems fine since being in `copy-mode-vi` is inherently a tmux interaction, so having the direct tmux navigation mappings makes sense to me there.

Issue #177, opened 2017-04-23
Problems when using arrow keys


Apromixately:
Since I am using a special keyboard where the arrow keys are easy to reach I wanted to set everything up to just use C-Direction. That required this special workaround http://stackoverflow.com/questions/15445481/mapping-arrow-keys-when-running-tmux. Would be nice if you could mention that somewhere.

blueyed:
Feel free to create a PR.. :)

PR #178, opened 2017-04-24
Added troubleshooting info for special keys

Apromixately:
as requested here: https://github.com/christoomey/vim-tmux-navigator/issues/177

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,47 @@ Consider moving code from your shell's non-interactive rc file (e.g.,
 `~/.zshenv`) into the interactive startup file (e.g., `~/.zshrc`) as Vim only
 sources the non-interactive config.
 
+### It Doesn't Work with the arrow keys
+
+You might have to tell your tmux how to send arrow keys to vim and tell vim how to expect them.
+Change your tmux config and your .vimrc by adding the following parts.
+This goes in your .vimrc:
+``` vim
+if &term =~ '^screen' && exists('$TMUX')
+    set mouse+=a
+    " tmux knows the extended mouse mode
+    set ttymouse=xterm2
+    " tmux will send xterm-style keys when xterm-keys is on
+    execute "set <xUp>=\e[1;*A"
+    execute "set <xDown>=\e[1;*B"
+    execute "set <xRight>=\e[1;*C"
+    execute "set <xLeft>=\e[1;*D"
+    execute "set <xHome>=\e[1;*H"
+    execute "set <xEnd>=\e[1;*F"
+    execute "set <Insert>=\e[2;*~"
+    execute "set <Delete>=\e[3;*~"
+    execute "set <PageUp>=\e[5;*~"
+    execute "set <PageDown>=\e[6;*~"
+    execute "set <xF1>=\e[1;*P"
+    execute "set <xF2>=\e[1;*Q"
+    execute "set <xF3>=\e[1;*R"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -230,6 +230,50 @@ Consider moving code from your shell's non-interactive rc file (e.g.,
...
+    execute "set <F9>=\e[20;*~"
+    execute "set <F10>=\e[21;*~"
+    execute "set <F11>=\e[23;*~"
+    execute "set <F12>=\e[24;*~"
+endif
+```
+
+This goes in your `~/.tmux.conf`:
+``` tmux
+set-option -gw xterm-keys on
```
blueyed: This is the default with tmux 2.4 btw.

README.md
```diff
@@ -230,6 +230,50 @@ Consider moving code from your shell's non-interactive rc file (e.g.,
...
 
+### It Doesn't Work with the arrow keys
+
+You might have to tell your tmux how to send arrow keys to vim and tell vim how to expect them.
+Change your tmux config and your `~/.vimrc` by adding the following parts.
+
+This goes in your `~/.vimrc`:
+``` vim
+if &term =~ '^screen' && exists('$TMUX')
+    set mouse+=a
```
blueyed: Mouse settings are not really relevant, are they?

README.md
```diff
@@ -230,6 +230,50 @@ Consider moving code from your shell's non-interactive rc file (e.g.,
...
+    execute "set <xF3>=\e[1;*R"
+    execute "set <xF4>=\e[1;*S"
+    execute "set <F5>=\e[15;*~"
+    execute "set <F6>=\e[17;*~"
+    execute "set <F7>=\e[18;*~"
+    execute "set <F8>=\e[19;*~"
+    execute "set <F9>=\e[20;*~"
+    execute "set <F10>=\e[21;*~"
+    execute "set <F11>=\e[23;*~"
+    execute "set <F12>=\e[24;*~"
```
blueyed: This should not be necessary with the correct TERM/terminfo setup.

Apromixately: I don't really understand the interactions of all the terminal settings but apparently vim does things differently when TERM=screen but if you have other suggestions for what I could try to make it work I'm all ears. :)

Apromixately: Are you referring to the function keys specifically?

blueyed: No, keys in general.
How about using `TERM=tmux-256color` after all?

I have this in my tmux.conf:
```
if-shell '[ "$TERM" = "linux" ]' 'set -g default-terminal "tmux"' 'set -g default-terminal "tmux-256color"'
```

Apromixately: That doesn't fix the problem for me.

christoomey:
Hello @Apromixately, thanks for writing this up and apologies for my delayed response. In the end (and really sorry I didn't comment in early to prevent unneeded work here), but I think all of the issues you're summarizing are generic tmux + vim configurations / sticking points, not anything specific to this plugin.

While this can certainly be useful information, my main goal with this plugin is to keep it working, and as simple and focused as possible. Namely, this plugin adds support for common navigation key bindings between Vim and tmux. Anything else is beyond the scope of this plugin.

Again, I want to apologize for not commenting in sooner, and to thank you for taking the time write this up. Even though I'm not merging it in, I really do appreciate you putting in the effort. I'm going to close this now based on the above summary, but please feel free to re-open or comment if you think I missed anything.

PR #179, opened 2017-06-05
fix the binding for the c-\

sjkyspa:
the C-\ is mis configured


blueyed:
Should also go to the README then, since it fixes https://github.com/sjkyspa/vim-tmux-navigator/commit/b93e176b7150942d9f0556347d1c8f6d0b690124.

christoomey:
@sjkyspa Thanks for pointing this out! We've merged in https://github.com/christoomey/vim-tmux-navigator/pull/181 which also provides a fix for this, so I'm going to close this now.

PR #181, opened 2017-06-19
properly shell escape backslash, fixes #180

hwine:

PR #182, opened 2017-07-07
Revert extra shell escaping in non-TPM bindings

christoomey:
Why
---

https://github.com/christoomey/vim-tmux-navigator/pull/181 was
introduced to fix an issues with the TPM install script not running
cleanly. This was due to the `C-\` key binding not being properly
escaped in the final line of the script.

In the process of fixing this in #181, it looks like we took it a bit
overboard and escaped _all_ `C-\` sequences. In the end, the others
should not be escaped, so this was causing issues for users not
configuring via TPM script.

This change
-----------

This change reverts all of the `C-\` escaping outside of the TPM script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,12 +64,12 @@ bind-key -n C-h if-shell "$is_vim" "send-keys C-h"  "select-pane -L"
 bind-key -n C-j if-shell "$is_vim" "send-keys C-j"  "select-pane -D"
 bind-key -n C-k if-shell "$is_vim" "send-keys C-k"  "select-pane -U"
 bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "select-pane -R"
-bind-key -n C-\\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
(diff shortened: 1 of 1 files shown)


christoomey:
@blueyed @hwine I believe my interpretation as stated in the PR description is accurate, but wondering if I might be missing anything. Any thoughts?

hwine:
I'm not quite clear on what you're viewing as the "non TPM script" installation. IIRC, I had the same 'unquoted backslash' problem when I copy/pasted from the README. So maybe there are 3 ways to install?

That said, since these are all README changes, I'm fine with them. Sorry to have left debris  behind.

christoomey:
The TPM script is evaluated as a shell script, whereas everything in the README is intended to be copied into the `~/.tmux.conf` file. As such, it seems that the TPM script installation requires the escaping, while the snippets intended for `~/.tmux.conf` usage do not. I've tested this locally and using `C-\\` does in fact error out, so I'm guessing you might have only used the TPM script? I am only aware of these two installation / configuration methods, so I think we should be all set now.

No worries about any debris, and thanks for providing the original fix! That was indeed broken, and now is not. And hopefully with this change, now all things are not broken!

PR #185, opened 2017-10-12
Add ability to extend matching pattern

bagrat:
As described in the **Nesting** section, one might want to extend the process name matching pattern, in order to capture `ssh`, `mosh` or `vagrant` in case of nesting.

To make that option easier, let's provide and environment variable to enable the extension with the default value set as the original one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,10 @@ the innermost tmux session and the vim sessions within that one. This works
 better than the default behaviour if you use the outer Tmux sessions as relays
 to different hosts and have all instances of vim on remote hosts.
 
+To easily extend the default pattern, you may use the `$VIM_TMUX_NAV_PAT`
+environment variable, and assign your custom pattern to it. To make it work,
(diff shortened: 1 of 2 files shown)


christoomey:
Hi @bagrat, thanks for sharing this. My initial inclination is that I'd prefer to keep the TPM install as simple and configuration free as possible, but I'd like to think on this a bit more. I'll reply back with thoughts in a few days.

christoomey:
Hi @bagrat, thanks again for sharing this. After thinking about it for a bit, I'm inclined to stick with my initial reaction and not merge this, with the suggestion that this kind of customization be relegated to the traditional copy & paste into `~/.tmux.conf` version of the setup, leaving the TPM install path with no configuration. I'm going to close this now based on the above summary, but please feel free to comment or re-open if you think I missed anything.

bagrat:
Hey @christoomey, thanks for the summary. I totally understand so no further comments. Cheers!

PR #192, opened 2017-12-23
Handle switching in insert mode

harterrt:
Currently, one cannot navigate to another pane if you're in vim in insert mode. This change adds additional `inoremap` to handle this case.

Criticism welcome.


christoomey:
Hi @harterrt, while I appreciate the effort I don't think this is a direction we want to go. In general I favor the "do as little in insert mode as possible", and I also think plugins should claim as few mappings as possible. Thankfully, as your PR shows this is entirely available to the end user to map on their own.

harterrt:
Cool, thanks for the review and happy holidays. :slightly_smiling_face:

PR #196, opened 2018-03-06
add: resize your vim windows like you will resize your tmux panes

dihydrro:
Hi,

i like the way tmux resize panes, so i took your plugin to simulate this behavior into vim and make it work along with tmux.

You can add this to your vimrc to try it:
```
nnoremap <silent> <C-H> :TmuxResizeLeft<return>
nnoremap <silent> <C-J> :TmuxResizeDown<return>
nnoremap <silent> <C-K> :TmuxResizeUp<return>
nnoremap <silent> <C-L> :TmuxResizeRight<return>
```

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -113,6 +113,54 @@ function! s:VimNavigate(direction)
   endtry
 endfunction
 
+function! s:TmuxResizeCmd(direction)
+  if s:InTmuxSession()
+    call s:TmuxAwareResize(a:direction)
+  else
+    call s:VimResize(a:direction)
+  endif
+endfunction
+
+" if we can't resize vim, we let tmux deal with it
+function! s:TmuxAwareResize(direction)
+  if !s:VimResize(a:direction)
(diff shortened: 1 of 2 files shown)


christoomey:
Hi @dihydrro, thanks for sharing this. This is certainly a novel addition to the existing functionality in the plugin. That said, I'm hesitant to add anything new to this project as my main goal is to keep it as simple and focused as possible. In addition, while navigation is largely unambiguous, I don't think resizing can be boiled down as simply.

I often find myself wanting to resize either the current vim window _or_ the tmux pane vim is within, and as such I would not want to blur the distinction there as I do with navigation.

Based on the reasons listed above I'm going to close this now, but thanks again for sharing.

dihydrro:
Thanks for the quick answer, I'm not surprise since this is a navigator plugin.

And thanks a lot for the plugin, its real helpful.

evs-chris:
Sorry to necro-comment here, but I stumbled across this while looking for the same functionality. So I thought I'd mention that I've forked this lovely plugin and used its bones to make a [vim-tmux-resizer](https://github.com/evs-chris/vim-tmux-resizer) plugin.

christoomey:
@evs-chris Never heard of "necro-comment"ing before 😀, but love the addition. Thanks for sharing!

PR #198, opened 2018-04-11
Change behaviour of <ctrl-{h,j,k,l}> only when necessary

urbainvaes:


christoomey:
Hey @urbainvaes, thanks for sharing these.

It looks like these new format arguments were only added in tmux 2.6. At a minimum I'd want to prevent this from breaking in tmux versions 2.5 and before, but I'm concerned about the complexity that would entail.

Can you describe the issue this change is working around so I can understand it better?

urbainvaes:
Hi! Thanks for the prompt response! 

Basically, this prevents `tmux` from capturing the <ctrl>-{h,j,k,l} keys when the `tmux` window is not splitted, or when there is no split to move to. This is especially useful to keep the readline mappings <ctrl-{h,k,l}> whenever possible.

PR #200, opened 2018-04-30
Refactor s:TmuxCommand, add s:GetTmuxCommand

blueyed:
- use a list instead of single string
- add s:GetTmuxCommand to be used later when not using `system()` only
- remove `:silent` when calling tmux from `s:TmuxAwareNavigate`: it
  should not be necessary and is bad practice to use `:silent`
  unnecessarily.  It was added in b068a04 with no explanation.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -28,7 +28,7 @@ function! s:InTmuxSession()
 endfunction
 
 function! s:TmuxVimPaneIsZoomed()
-  return s:TmuxCommand("display-message -p '#{window_zoomed_flag}'") == 1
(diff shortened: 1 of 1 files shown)


christoomey:
Broadly these changes all look good to me, but unfortunately I'm getting an error when running with this locally:

```
Error detected while processing function <SNR>151_TmuxWinCmd[2]..<SNR>151_TmuxAwareNavigate[23]..<SNR>151_TmuxCommand:
line    2:
E730: using List as a String
```

Any thoughts on what the issue might be?

blueyed:
@christoomey 
The problem is that Vim's `system()` does not handle lists for the command/argv.  Will fix it.

PR #219, opened 2018-11-02
Remove README section about Neovim incompatibility

goronfreeman:
I've been using Neovim with tmux-navigator with no issues for many months now. I just noticed that the README calls out an incompatibility with Neovim and links to an issue. The issue was called out as fixed in [this](https://github.com/neovim/neovim/issues/2048#issuecomment-269032502) comment.

I think this section can safely be removed from the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,8 +247,8 @@ sources the non-interactive config.
 
 ### It doesn't work in Vim's `terminal` mode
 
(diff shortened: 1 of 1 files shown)


christoomey:
Thanks for update @goronfreeman!

PR #221, opened 2018-11-22
Take window boundaries into account

svanharmelen:
I personally like this behavior very much, but if you think otherwise could we then maybe support is behind a configuration option?

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -108,7 +108,17 @@ function! s:TmuxAwareNavigate(direction)
       catch /^Vim\%((\a\+)\)\=:E141/ " catches the no file name error
       endtry
     endif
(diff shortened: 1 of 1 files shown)


christoomey:
Thanks for sharing this @svanharmelen. If I'm reading correctly this will make the navigation keys do nothing if at the relevant edge of the screen, as opposed to wrapping around as they currently do? I've not heard much from folks looking for this functionality, and I personally wouldn't want this, so I'd at least want to see this behind an option. That said, it also only works on versions of tmux from 2.6 and up. All that combined leans me towards closing this as my driving goal here is to keep the plugin as simple as possible, and make it largely "just work" for folks. Again, thanks for sharing this, but for now I'm going to close this.

svanharmelen:
All good! Thanks for the feedback 👍🏻

PR #225, opened 2019-01-08
Mention that tmux must be present in PATH

nmattia:
This plugin didn't work for me, commands wouldn't do anything. E.g. `:TmuxNavigateRight` would simply be a no-op. When running `:TmuxNavigatorProcessList` I got an error:

> /bin/bash: tmux: command not found

After [adding `tmux` to the PATH](https://github.com/nmattia/homies/commit/06aa54743990613f3b53a3bf7334d23ce59acc4a) everything worked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ configuration described in [this gist][]. When combined with a set of tmux
 key bindings, the plugin will allow you to navigate seamlessly between
 vim and tmux splits using a consistent set of hotkeys.
 
-**NOTE**: This requires tmux v1.8 or higher.
+**NOTE**: This requires tmux v1.8 or higher and tmux must be present in Vim's `PATH`.
 
 Usage
 -----


christoomey:
Hi @nmattia, thanks for sharing this. While I appreciate the update, I want to balance the amount of detail in the README and my gut says this one is outside the threshold of utility. That said, I think this issue can serve equally well if not better due to being searchable, but I'm wondering if you wouldn't mind adding a few more notes to help clarify things in case anyone finds this issue down the road:

1. Can you share the specific error message you hit?
2. Can you clarify what you did to get `tmux` on your `$PATH`?

I'm going to close this now, but if you're up for it please do comment back with the answers to those questions so that we can have a bit more detail / grepability here.

nmattia:
Hi @christoomey , makes sense. Maybe worth adding it to the [README's troubleshooting](https://github.com/christoomey/vim-tmux-navigator#troubleshooting) section?

1. When trying to change tmux panes with Ctrl-(hjkl) from within Vim, nothing at all happens. Same when calling `:TmuxNavigateRight`. When calling `:TmuxNavigatorProcessList`, I get the following error: `/bin/bash: tmux: command not found`
1. How to add tmux to vim's `PATH` is very much system dependent. If vim inherits the user's `PATH` chances are `tmux` is already present. In my case I had conflicting python environments so I had to [overwrite vim's `PATH`](https://github.com/nmattia/homies/blob/master/vim/default.nix#L38-L40), which hid `tmux` from it. To fix the issue I just re-exported tmux's path in vim's `PATH`.

PR #229, opened 2019-02-05
adds fzf compatibility (in shell)

proever:
this adds the ability to use C-j and C-k to navigate between entries that FZF has generated, for example when using C-r to find history items in the shell. This does not address the incompatibility with fzf in vim's term mode addressed on the plugin's readme.

### DIFF
--- a/vim-tmux-navigator.tmux
+++ b/vim-tmux-navigator.tmux
@@ -2,9 +2,11 @@
 
 is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
     | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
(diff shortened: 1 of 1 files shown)


proever:
the `$is_fzf` snippet was taken from https://blog.bugsnag.com/tmux-and-vim/

christoomey:
Hi @proever, I appreciate you taking the time to create this PR, but I'm inclined to not include this in the core setup instructions. My overarching goal here is to keep things as simple (and supportable 😄) as possible, and as such I'm very cautious with what I add to the core configuration logic. Based on that I'm going to close this as it stands, but I think it could make a great addition as a section specific to FZF lower down in the README if you felt up to tackling that.

proever:
Ok, no worries. Thanks for taking the time to look. When you say a section in the README do you mean something that explains how to modify the plugin yourself in the way I did in 351c046211889ab3e221118488380f53c188c169?

PR #231, opened 2019-04-15
Add support for forwarding commands to an external script.

tafryn:
This lets you incorporate window manager navigation on the same set of keybindings that this plugin uses to navigate between tmux and vim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,17 @@ With this enabled you can use `<prefix> C-l` to clear the screen.
 
 Thanks to [Brian Hogan][] for the tip on how to re-map the clear screen binding.
 
+#### Forwarding Commands to External Script
+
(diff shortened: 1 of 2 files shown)


christoomey:
Hey @tafryn, thanks for sharing this. While I sincerely appreciate the effort you've put in here, I tend to be very conservative in what I incorporate in this plugin as my main goal is to keep it as simple and maintainable as possible. You're the first person I've seen with this concern, and frankly interacting with something other than tmux makes me feel like we're trying to force fit this plugin to cover more cases than it should.

With all that in mind, I'm going to close this. Again, I really do appreciate the PR and I hope the above reasoning makes sense. Please feel free to comment or reopen if you think I missed something.

Issue #233, opened 2019-06-21
Add option to prevent wrap around pane movement


himat:
Is it possible to add an option to this plugin to prevent `tmux select-pane` from being called if the current vim window is on the edge of the screen?
So that if I'm in the right-most tmux pane with a vim window inside it, then pressing `ctrl-l` won't wrap around to then going to the left-most pane.

Tmux doesn't have an option to do this, so I added my own command to do so.
I was able to stop the wrap-around on the tmux side by putting this line in my tmux.conf
`bind -n C-l run-shell "if [ $(is_not_rightmost) ]; then tmux select-pane -R; fi"` so that it only selects the next right pane if it's not at the right edge of the screen.
(where `is_not_rightmost="$(tmux display-message -p '#{pane_at_right}') -ne 1"`)

And then to use it with vim-tmux-navigator, I would change that line above to be
`bind -n C-l run-shell "if $is_vim; then tmux send-keys C-l; elif [ ${is_not_rightmost} ]; then tmux select-pane -R; fi"`

But I would need support from within this plugin to make it work if a vim window is open in a pane at the edge of the screen. Since I send-keys to the plugin, which then changes the tmux pane internally.

PR #237, opened 2019-07-17
Add option and instructions to disable wrapping

heewa:
* Add a vim option (`g:tmux_navigator_no_wrap`) to disable wrapping from within vim
* Add instructions to the Readme for configuring tmux bindings to only select panes when it wouldn't wrap
* Selecting previous tab is unaffected

Addresses #233

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,15 +221,15 @@ to conditionally wrap based on position on screen:
 # See: https://github.com/christoomey/vim-tmux-navigator
 is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
     | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
-bind-key -n C-h if-shell "$is_vim" "send-keys C-h"  "if -F '#{?pane_at_left,0,1}' 'select-pane -L'"
-bind-key -n C-j if-shell "$is_vim" "send-keys C-j"  "if -F '#{?pane_at_bottom,0,1}' 'select-pane -D'"
-bind-key -n C-k if-shell "$is_vim" "send-keys C-k"  "if -F '#{?pane_at_top,0,1}' 'select-pane -U'"
-bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "if -F '#{?pane_at_right,0,1}' 'select-pane -R'"
+bind-key -n C-h if-shell "$is_vim" "send-keys C-h"  "if-shell 'expr #{pane_left}' 'select-pane -L'"
+bind-key -n C-j if-shell "$is_vim" "send-keys C-j"  "if-shell 'expr #{window_height} - #{pane_bottom} - 1' 'select-pane -D'"
+bind-key -n C-k if-shell "$is_vim" "send-keys C-k"  "if-shell 'expr #{pane_top}' 'select-pane -U'"
+bind-key -n C-l if-shell "$is_vim" "send-keys C-l"  "if-shell 'expr #{window_width} - #{pane_right} - 1' 'select-pane -R'"
 bind-key -n C-\ if-shell "$is_vim" "send-keys C-\\" "select-pane -l"
-bind-key -T copy-mode-vi C-h if -F '#{?pane_at_left,0,1}' 'select-pane -L'
-bind-key -T copy-mode-vi C-j if -F '#{?pane_at_bottom,0,1}' 'select-pane -D'
-bind-key -T copy-mode-vi C-k if -F '#{?pane_at_top,0,1}' 'select-pane -U'
-bind-key -T copy-mode-vi C-l if -F '#{?pane_at_right,0,1}' 'select-pane -R'
+bind-key -T copy-mode-vi C-h if-shell 'expr #{pane_left}' 'select-pane -L'
+bind-key -T copy-mode-vi C-j if-shell 'expr #{window_height} - #{pane_bottom} - 1' 'select-pane -D'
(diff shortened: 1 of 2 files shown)


christoomey:
I believe this is similar to what was presented in https://github.com/christoomey/vim-tmux-navigator/pull/221 and my concerns, most notably around backward compatibility, still hold. Would it be possible to wrap this in a version check ensuring we have a sufficiently recent tmux?

heewa:
Great call out!

I tried a few different things, and the best solution I landed on is using the portable `expr` command with tmux variables that exist in older versions, all the way back to 2.0 - see [`format.c` from the 2.0 tag](https://github.com/tmux/tmux/blob/2.0/format.c#L676-L679)).

Tested with 2.8 & 2.5 tmux, with wrap & no-wrap settings, regular panes & panes with vim (with the option enabled & disabled). Both wrap & no-wrap work in old & new tmux.

I didn't test with a non-bash shell, but I'm not using any bash syntax, so it should work in any shell, as long as the `expr` command is available. From light googling around, I think that should be pretty portable. If you know of a more cross-shell & cross-distro approach, I'll switch to that.

heewa:
Also, to be clear, if this doesn't work on someone's setup for whatever reason, they could just not enable the option, and this code path doesn't get hit. The plugin functions normally, with regular wrapping behavior for those people.

nickspoons:
This PR is exactly what I was looking for. It works very well for me, thanks @heewa

Asocia:
Thank you very much for this! I was going to try to fix this myself but then I thought maybe somebody already did this :) I love the power of community!

README.md
```diff
@@ -202,6 +202,38 @@ bind -r C-l run "tmux select-pane -R"
...
+
+Tmux doesn't have an option, so whatever key bindings you have need to be set
+to conditionally wrap based on position on screen:
+``` tmux
+# Smart pane switching with awareness of Vim splits.
+# See: https://github.com/christoomey/vim-tmux-navigator
+is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
+    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
+bind-key -n C-h if-shell "$is_vim" "send-keys C-h"  "if-shell 'expr #{pane_left}' 'select-pane -L'"
+bind-key -n C-j if-shell "$is_vim" "send-keys C-j"  "if-shell 'expr #{window_height} - #{pane_bottom} - 1' 'select-pane -D'"
```
maximkulkin, line 225: All checks can be unified into same structure:
```
bind-key -n C-h if-shell "$is_vim" { send-keys C-h } { if -F '#{pane_at_left}'   {} { select-pane -L } }
bind-key -n C-j if-shell "$is_vim" { send-keys C-j } { if -F '#{pane_at_bottom}' {} { select-pane -D } }
bind-key -n C-k if-shell "$is_vim" { send-keys C-k } { if -F '#{pane_at_top}'    {} { select-pane -T } }
bind-key -n C-l if-shell "$is_vim" { send-keys C-l } { if -F '#{pane_at_right}'  {} { select-pane -R } }
```

plugin/tmux_navigator.vim
```diff
@@ -109,6 +113,18 @@ function! s:TmuxAwareNavigate(direction)
       endtry
     endif
     let args = 'select-pane -t ' . shellescape($TMUX_PANE) . ' -' . tr(a:direction, 'phjkl', 'lLDUR')
+    " Prevent wrapping, if option is set, by checking if last pane in appropriate direction
+    if g:tmux_navigator_no_wrap == 1
+      if a:direction == 'h'
```
maximkulkin, line 118: This can be condensed to just two lines using dictionaries:
```
let positions = {'h': 'left', 'j': 'bottom', 'k': 'top', 'l': 'right'}
let args = 'if -F "#{pane_at_' . positions[a:direction] . '}" "" "' . args . '"'
```
the definition for dictionary can then be moved to some global constant

README.md
```diff
@@ -202,6 +202,38 @@ bind -r C-l run "tmux select-pane -R"
...
+
+For vim, you only need to enable this option:
+``` vim
+let g:tmux_navigator_no_wrap = 1
+```
+
+Tmux doesn't have an option, so whatever key bindings you have need to be set
+to conditionally wrap based on position on screen:
+``` tmux
+# Smart pane switching with awareness of Vim splits.
```
maximkulkin, line 220: This really should also go into [vim-tmux-navigator.tmux](https://github.com/christoomey/vim-tmux-navigator/blob/master/vim-tmux-navigator.tmux)

maximkulkin:
@christoomey Hey Chris, what do we need to make it happen?

christoomey:
Hey @maximkulkin @heewa, thanks for the work here and apologies for my delayed replies.

Looking at the way you've implemented things I feel comfortable merging this in as it's an opt-in enhancement for folks who want it but doesn't complicate the base functionality (thanks for that!).

It does look like there's a merge conflict in the `plugin/tmux_navigator.vim` file. I took a quick look but it seems like a potentially real conflict rather than just adjacent lines, so I'm hoping one of you can take a look and rebase / resolve as needed. Once we have that, I'm happy to merge 🙂

maximkulkin:
In case original author gave up on it, I have created an updated PR #320

christoomey:
This has been merged via #320. Thanks for the work @heewa and sorry for the _long_ delay in getting this in, and thanks to @maximkulkin for getting this to the finish line.

PR #244, opened 2019-12-02
Add warning about changes to tmux v3++ configuration

willfindlay:
Added a small message to README about needing to use `'C-\'` in new versions of tmux instead of `C-\` for tmux configuration. Related to https://github.com/tmux/tmux/issues/1827

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ bind-key -T copy-mode-vi C-l select-pane -R
 bind-key -T copy-mode-vi C-\ select-pane -l
 ```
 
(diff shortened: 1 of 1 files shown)


christoomey:
Thanks for the update, @willfindlay!

l-qing:
Only change `C-\` to `'C-\'`, It's tips "Syntax error" on use.
Also need to be modified `"send-keys C-\\"` to `'send-keys C-\\'` or `"send-keys C-\\\\"`

Like this:
`bind-key -n 'C-\' if-shell "$is_vim" 'send-keys C-\\' "select-pane -l"`

PR #263, opened 2020-06-05
Fix vim pack README.md incorrect directory

cevans87:
On Ubuntu 20.04 with tmux 3.0a and vim 8.1.2269

Navigation from vim -> tmux did not work. Vim plugin directory layout expects a `config` subdirectory after the `start` directory. The fixed lines point to the correct location to clone this repo so that vim will load the plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,10 +47,10 @@ Then run
 :PluginInstall
 ```
 
-If you are using Vim 8+, you don't need any plugin manager. Simply clone this repository inside `~/.vim/pack/plugin/start/` directory and restart Vim.
+If you are using Vim 8+, you don't need any plugin manager. Simply clone this repository inside `~/.vim/pack/plugin/start/config/` directory and restart Vim.
 
 ```
-git clone git@github.com:christoomey/vim-tmux-navigator.git ~/.vim/pack/plugin/start/
+git clone git@github.com:christoomey/vim-tmux-navigator.git ~/.vim/pack/plugin/start/config/
 ```
 
 


README.md
```diff
@@ -47,10 +47,10 @@ Then run
 :PluginInstall
 ```
 
-If you are using Vim 8+, you don't need any plugin manager. Simply clone this repository inside `~/.vim/pack/plugin/start/` directory and restart Vim.
+If you are using Vim 8+, you don't need any plugin manager. Simply clone this repository inside `~/.vim/pack/plugin/start/config/` directory and restart Vim.
```
christoomey, line 50: Hmm, I'll be honest that I'm still using vim-plug so I'm less well versed with the new native package manage stuff, but this doesn't look quite right to me. Based on [these docs](https://vimhelp.org/repeat.txt.html#package-create), it looks like perhaps we're missing a top-level `vim/pack/start/vim-tmux-navigator` directory or similar. Mind taking a look at that?

Luiserebii:
I was having this issue, and was struggling for a bit, but following what this pull request suggested with the directory worked perfectly for me. For reference, I'm on tmux 3.1b, and vim 8.1.

jsrozner:
I actually had a very similar issue, but my solution was to put this into ˜/.vim/pack/bundle/start

I also was unable to use vim -> tmux (and also unable to do vim -> vim). I followed the instructions here: https://vi.stackexchange.com/questions/613/how-do-i-install-a-plugin-in-vim-vi

Which indicate that we should use ˜/.vim/pack/bundle/start/<plugin files>

This fixed the problem for me.

I spent about an hour trying to figure out why this package was not working.

README.md
jsrozner: See https://vi.stackexchange.com/questions/613/how-do-i-install-a-plugin-in-vim-vi

It seems like we should actually be doing s/plugin/bundle

This change worked for me. See also my comment below.

christoomey:
I think this has been fixed by another PR which was just merged, so I'm going to close this, but please feel free to comment or re-open if I'm missing anything.

jsrozner:
I think it should still be ˜/.vim/pack/bundle/start/ rather than ~/.vim/pack/plugins/

In your readme, it should be a change to this: 

> If you are using Vim 8+, you don't need any plugin manager. Simply clone this repository inside ~/.vim/pack/plugin/start/ directory and restart Vim.
> 
> git clone git@github.com:christoomey/vim-tmux-navigator.git ~/.vim/pack/plugins/start/vim-tmux-navigator
> 

I think we should switch plugin -> bundle in both of these strings

i.e., "clone this repo inside `~/ .vim / pack / bundle / start / <github repo>`
and `git clone git@github.com:christoomey/vim-tmux-navigator.git ~/.vim/pack/bundle/start/vim-tmux-navigator`

christoomey:
Reading the docs and examples, I _think_ the directory under `~/.vim/pack/<dir>` can be named anything you want, and there can even be multiple. I just did a quick test with plugins in both `~/.vim/pack/bundle/start/` and `~/.vim/pack/plugins/start/` and both loaded as expected. I'm also more inclined to stick with the `pack`/`plugin` naming than `bundle` since that comes from Vundle and the like. So all total, I think we're still good here, but thanks again for the notes and effort.

PR #266, opened 2020-06-05
Don't shell out when detecting vim

joshfrench:
Even with a totally bare shell, my tmux lags when calling `if-shell`. This PR uses tmux's `if-shell -F` and regex format to test `#{pane_current_command}`, returning true/false without opening a shell at all.

I'm not sure what it loses by using `pane_current_command` instead of `ps` & `pane_tty` so this may be a portability regression, but I wanted to share in case anyone else is experiencing shell-related slowness. Perhaps someone who has more context on why `ps` is preferable can cobble together an implementation that is both portable and shell-free?

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,12 +65,11 @@ Add the following to your `~/.tmux.conf` file:
 ``` tmux
 # Smart pane switching with awareness of Vim splits.
 # See: https://github.com/christoomey/vim-tmux-navigator
-is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
-    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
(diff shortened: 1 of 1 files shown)


christoomey:
Hello @joshfrench, thanks for sharing this. Unfortunately switching back to using `pane_current_command` would be a regression around some use cases we want to support, e.g. cases where vim has a subprocess running (see [this issue](https://github.com/christoomey/vim-tmux-navigator/issues/96) for more detail).

Anecdotally, I've not heard of others experiencing this slowdown, and in my setup I'm able to move between panes with essentially no lag (despite having a 1-2 second shell startup time), so I wonder if something else is at play here for you?

Thankfully you can make this configuration in your setup so I don't think there is anything to do in the documentation proper, so with that in mind I'm going to close this issue, but please feel free to comment or reopen if you think I missed something here.

joshfrench:
No worries, I figured this was a non-starter. Thanks for the context on using `ps`!

PR #268, opened 2020-06-10
Added option g:tmux_navigator_disable_wrapping

cpakkala:
Added an option to prevent wrapping top/bottom or left/right when trying to move in a direction in which no tmux pane exists.  By default (if the new option is not set) the prior behavior of wrapping still occurs.


cpakkala:
Ahh, crap, my bad.  I totally missed #237 addressing this already...

PR #269, opened 2020-06-21
Add tmux executable option

hnspn:
Adding the `g:tmux_executable` option.

My tmux is not installed globaly so I need this.

Maybe some other people need this.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -46,8 +46,12 @@ if !exists("g:tmux_navigator_disable_when_zoomed")
   let g:tmux_navigator_disable_when_zoomed = 0
 endif
 
+if !exists("g:tmux_executable")
+  let g:tmux_executable = 'tmux'
+endif
+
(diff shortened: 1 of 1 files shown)


christoomey:
Hey @hnspn, I'm happy to consider this addition, but first I'm wondering if there might be other options that you could take that wouldn't require changes to the plugin code.

Specifically, I'm wondering if you can either:

1. Assuming your executable is just installed in a non-standard path, add the tmux executable location to your `$PATH`, e.g. `export PATH=$PATH:/new/path/where/tmux/is` (assuming tmux is in the `/new/path/where/tmux/is` directory)
2. If your executable is named something different, could you alias it, e.g. `alias tmux=/other/path/tmux-like-executable`

My guess is that between those or similar options you might be able to solve this entirely within your configuration, and I'm always inclined to do that rather than add more code here.

hnspn:
You are right.
Changing my path works.
The alias does not.
Will close.

PR #276, opened 2020-09-14
Fix vim-tmux-navigator version checks

matthewstrasiotto:
New pattern tested on "tmux 3.0a", and "tmux next-3.3" . 

Resolved issue with TPM script that  was(incorrectly) directly calling `if-shell`, (as though it was a tmux source script). 

TPM scripts are actually bash scripts, that invoke tmux bindings through the tmux executable, and the correct behaviour is to call `tmux if-shell`.


christoomey:
Thanks for the update here! I'm not a user of TPM myself so I'll have to trust that this works across versions, but given that things are known to be broken for pre-release use cases, this feels like a worthwhile update.

PR #281, opened 2020-09-29
temporarily set shellcmdflag to -c before calling system

mmjo:
When i had shellcmdflag set to '-ic' tmux navigation failed.
This patch fixes that.

Regards, Mike


christoomey:
Thanks for the update!

PR #296, opened 2021-05-31
Add DownOrRight and UpOrLeft directions.

modk:
From other tools I am used to switching windows/panes in a CCW way
where an 'up' translates to a 'left' if there is no window above and
a 'down' becomes a 'right' if there is no window below.

This patch adds two new commands

 `TmuxNavigateDownOrRight`
 `TmuxNavigateUpOrLeft`

that will check check for a neighbor in the requested direction and,
if there is none, change the direction.

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -7,9 +7,51 @@ if exists("g:loaded_tmux_navigator") || &cp || v:version < 700
 endif
 let g:loaded_tmux_navigator = 1
 
-function! s:VimNavigate(direction)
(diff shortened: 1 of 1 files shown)


christoomey:
Hi @modk, while I really appreciate you sharing this, I'm inclined to not merge this into the core. My guiding principle with this plugin is to keep things as stable as possible (ideally "it just works" for most users), and as an extension of this, I try to keep things as simple as possible. While the functionality you're suggesting here seems interesting, I'm not comfortable taking on the added code complexity needed to support it. Hopefully this PR can serve as a reference if any is looking for this functionality down the road, and sincerely, thanks again for sharing this.

Issue #295, opened 2021-05-28
`is_vim` does not detect vim running in a subshell


jyurek:
I have a project where I am working in Python. I'm using `pipenv` to manage the environment, which has a feature where if you run `pipenv shell` it starts a subshell with everything set up right for the environment. This starts a new TTY.

However, that means that `ps ... -t "#{pane_tty}"` returns the "wrong" process list. It returns 
```
  PID TTY           TIME CMD
92716 ttys001    0:00.85 -bash
96594 ttys001    0:22.72 [/path/to/]Python /usr/local/Cellar/pipenv/2020.11.1
```

But just `ps` alone returns (and you can see `vim` running inside `/dev/ttys002`)
```
  PID TTY           TIME CMD
14445 ttys000    0:00.29 -bash
14803 ttys000    0:00.00 /bin/sh /Users/jyurek/bin/tm [my_project]
14804 ttys000    0:00.01 tmux new-session -As [my_project]
92716 ttys001    0:00.85 -bash
96594 ttys001    0:22.72 [/path/to/]Python /usr/local/Cellar/pipenv/2020.11.1
53522 ttys002    0:19.13 vim [source_file].py
96606 ttys002    0:01.01 /bin/bash -i
 8581 ttys003    0:00.33 -bash
61829 ttys003    0:07.61 [/path/to/]Python /usr/local/Cellar/pipenv/2020.11.1
61835 ttys005    0:04.68 /bin/bash -i
```

The end result is that the `is_vim` clause in the pane-switching bindings doesn't return true, so it doesn't send the keys to vim, even though it's the program in the foreground.

I'm not really sure how to fix this, since apparently OSX's `ps` can't show the lineage of processes to know that `vim` is, in fact, running as an (eventual) sub-process in `/dev/ttys001`.

christoomey:
Hey John 👋. Well, this is certainly a fun one. I'll be honest that I don't have a great grasp of the `ps` mechanics at play here (those tweaks were added by a contributor), but I think you might be able to tweak things on your end to get a working solution. The subtlety lies in avoiding false positives so I'm always reticent to change the vim matching logic, but here's two suggestions:

1. Try out [Mislav's original script](https://gist.github.com/mislav/5189704) (I wouldn't be surprised if this didn't work either as it's purposefully a bit less thorough in how it tries to match processes, but could be an option)
[...]

PR #300, opened 2021-07-30
Update `is_vim` detection logic for tmux 3.2

christoomey:
Users upgrading to tmux 3.2 have been reporting issues with using the
plugin, specifically that it would "freeze" their terminal for 30+
seconds.

@kidonchu kindly provided an updated version of the `is_vim` check that
seems to resolve the issue, so this PR updates the README and TPM script
to use that version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,7 @@ Add the following to your `~/.tmux.conf` file:
 ``` tmux
 # Smart pane switching with awareness of Vim splits.
 # See: https://github.com/christoomey/vim-tmux-navigator
-is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
-    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
+is_vim="ps -o state=,tty=,comm= | grep -iqE '^[^TXZ ]+ +#{s|/dev/||:pane_tty}\s+(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
 bind-key -n 'C-h' if-shell "$is_vim" 'send-keys C-h'  'select-pane -L'
 bind-key -n 'C-j' if-shell "$is_vim" 'send-keys C-j'  'select-pane -D'
 bind-key -n 'C-k' if-shell "$is_vim" 'send-keys C-k'  'select-pane -U'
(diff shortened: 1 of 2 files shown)


dkarter (approved):
Looks good, thank you for the fix! Tested with tmux next-3.3 and Neovim 0.5.

christoomey:
@dkarter thanks for the note! I'm also looking to confirm if this version of the snippet is reasonably backwards compatible, so hoping to hear from folks on tmux < 3.2 to validate that, so will likely need to keep this open for a bit.

mhartington:
👋 Hello! So I gave this a shot on my setup with tmux 3.2, but I think the issue is still there.

```
  is_vim="ps -o state=,tty=,comm= | grep -iqE '^[^TXZ ]+ +#{s|/dev/||:pane_tty}\s+(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
  bind -n C-h if-shell "$is_vim" "send-keys C-h" "select-pane -L"
  bind -n C-j if-shell "$is_vim" "send-keys C-j" "select-pane -D"
  bind -n C-k if-shell "$is_vim" "send-keys C-k" "select-pane -U"
  bind -n C-l if-shell "$is_vim" "send-keys C-l" "select-pane -R"
```


With this plus the typical vim keybindings, it seems to skip over vim splits and goes directly to chaning tmux panes.

unphased:
Hi could you clarify what `fig` is?

yut23:
This doesn't work for me on Fedora 34 (using ps from procps-ng 3.3.17), but passing `-a` to ps fixes it. Without that, it seems to just list all the processes with no connected terminal.

sullivan-sean:
Here's

> 👋 Hello! So I gave this a shot on my setup with tmux 3.2, but I think the issue is still there.
> 
> ```
>   is_vim="ps -o state=,tty=,comm= | grep -iqE '^[^TXZ ]+ +#{s|/dev/||:pane_tty}\s+(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
>   bind -n C-h if-shell "$is_vim" "send-keys C-h" "select-pane -L"
>   bind -n C-j if-shell "$is_vim" "send-keys C-j" "select-pane -D"
>   bind -n C-k if-shell "$is_vim" "send-keys C-k" "select-pane -U"
>   bind -n C-l if-shell "$is_vim" "send-keys C-l" "select-pane -R"
> ```
> 
> With this plus the typical vim keybindings, it seems to skip over vim splits and goes directly to changing tmux panes.
> 
> EDIT: Turns out fig was causing some issue where it modified the results from `TmuxNavigatorProcessList`. Uninstalled and everything worked fine.

I think this approach https://github.com/christoomey/vim-tmux-navigator/issues/295#issuecomment-922092511 should fix incompatibility with Fig

PR #304, opened 2021-09-19
Add option to avoid overriding the C-\ key.

bergercookie:
Add flag to avoid overriding the C-\ key

The `C-\` key is rather useful since it's normally mapped to the `SIGQUIT`
signal. This commit adds a flag to disable this mapping if the user wishes to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,15 @@ To disable navigation when zoomed, add the following to your ~/.vimrc:
 let g:tmux_navigator_disable_when_zoomed = 1
 ```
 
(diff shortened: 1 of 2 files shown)


christoomey:
Hi @bergercookie, appreciate the suggestion here, but I'd prefer to avoid the additional configuration here. The TPM download is meant to be the one liner simple approach to the installation, but for users who want more customization, e.g. avoiding this particular mapping, the default install notes w/ manually adding the lines to the tmux.conf should work fine.

PR #309, opened 2022-01-04
Run Neovim default <C-l> before TmuxNavigateRight

craigmac:
This PR detects when we are running Neovim, and if we are runs the
default <C-l> Neovim mapping first, then calls the `:TmuxNavigateRight`.

The purpose is to respect the default  Neovim <C-l> behaviour. Some may not like this as it is now doing two things: turning off the search highlighting and updating the diff, and then trying to move to the window to the right. Thoughts?

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -17,19 +17,14 @@ endfunction
 
 if !get(g:, 'tmux_navigator_no_mappings', 0)
   if has('nvim')
-    " Neovim should run its default <C-l> binding first
-    nnoremap <silent> <C-l> <Cmd>nohlsearch<Bar>diffupdate<Bar>TmuxNavigateRight<cr>
-    nnoremap <silent> <C-h> <Cmd>nohlsearch<Bar>diffupdate<Bar>TmuxNavigateLeft<cr>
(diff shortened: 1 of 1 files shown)


craigmac:
Sure, makes sense to me and it's what I did. In my vimrc I use this to backport the <C-l> default from Neovim (Neovim users can adjust for their init.vim to get it working with this plugin):

In `~/.vim/vimrc`:
 
```vim
let g:tmux_navigator_no_mappings=1
nnoremap <silent> <C-l> <Cmd>nohlsearch<Bar>diffupdate<Bar>TmuxNavigateRight<CR>
nnoremap <silent> <C-h> <Cmd>TmuxNavigateLeft<cr>
nnoremap <silent> <C-j> <Cmd>TmuxNavigateDown<cr>
nnoremap <silent> <C-k> <Cmd>TmuxNavigateUp<cr>
nnoremap <silent> <C-\> <Cmd>TmuxNavigatePrevious<cr>
```

craigmac:
> definitely surprised to hear it's neovim's default

To be fair, Neovim does call plain ol' `<C-l>` at the end of their remapping in Neovim, it's just that it won't work with this plugin that way because of script loading order.

PR #310, opened 2022-01-09
send-keys when there is only one pane

jaywonchung:
The idea is that when there's only one pane in the current tmux window, C-{h,j,k,l,\\} must be sent to the shell as is. For instance, without this, C-k (kill all characters after the cursor) and C-l (clear the screen) will not work inside tmux.

I'm a novice in tmux and making things compatible with all shells. If you agree with the idea, please feel free to edit this PR as you see fit.

### DIFF
--- a/vim-tmux-navigator.tmux
+++ b/vim-tmux-navigator.tmux
@@ -2,21 +2,22 @@
 
 version_pat='s/^tmux[^0-9]*([.0-9]+).*/\1/p'
 
-is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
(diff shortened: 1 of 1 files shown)


christoomey:
Hi @jaywonchung, thanks for sharing this. While I certainly appreciate the effort here, I tend to be quite conservative in what I add to the plugin as my core goal is to make it "just work" in as many situations as possible, and any added functionality / complexity makes that harder.

In this case, while the idea is certainly sound, I'm comfortable with [the other solutions we offer](https://github.com/christoomey/vim-tmux-navigator#restoring-clear-screen-c-l) as to how to work around the issues you highlighted.

With that in mind I'm going to close this, but please feel free to comment or reopen if you think I missed something.

PR #314, opened 2022-03-15
tmux-aware vim split-resizing

martin-louazel-engineering:
Hello everyone !

First of all, thanks a lot for the work you put on the plugin, I enjoy using it very much, and it is making my life quite easier !

I wanted to potentially improve on the plugin by also allowing to resize easily between vim splits and tmux panes.
The usual use-case would be this: When I am looking at something in vim, there are a lot of time where I want to make a split wider to get a bigger view. The problem is, whenever the split covers the full tmux pane, if I want it to take more space, I am usually stuck as far as vim commands go. What I usually have to do then is switch to a different tmux pane, do the resizing in tmux, and then go back, which I find quite frustrating every time. Plus, I always found the split-resizing in vim to be quite painful in general.

So if we could get a way to resize both tmux panes and vim split as if they were the same thing, that would kill 2 birds with one stone. And I think I have achieved that.

I have to say, I am absolutely no specialist neither in vimscript nor in developing vim plugins, so there are probably some obvious things that are not optimized, or that I forgot to add in, especially when it comes to other tmux/tmate/vim versions. These changes work without any problems (at least for the basic use that I have of vim and tmux) for vim v8.0 and tmux v2.6. 

Anyway, if you have any feedback on that, no matter whether or not you want to accept it, I would love to hear from you.

The problem space turned out to be surprisingly tricky, so it might not be 100% apparent what all the code does at first sight I admit. So if you have any question don't hesitate to ping me, I'll be very happy to answer them.

One thing that I think helped me a lot was to think about the tmux resizing behavior as "moving a separator/splitter" (instead of "growing the cell by X amount", like in vim), where the separator to be moved is always the one on the left/at the bottom, unless your at the left/bottom edge of the window, in which case it will be the separator directly on the right/top, respectively.
Long story short, a lot of what I am doing here is finding where the tmux separator is supposed to be compared to the current vim split.

So that's it, I hope you like it and find it somewhat interesting, if not useful.
Again, thanks a lot for your time and effort.
Bye !

### DIFF
--- a/plugin/tmux_navigator.vim
+++ b/plugin/tmux_navigator.vim
@@ -1,6 +1,8 @@
 " Maps <C-h/j/k/l> to switch vim splits in the given direction. If there are
 " no more windows in that direction, forwards the operation to tmux.
 " Additionally, <C-\> toggles between last active vim splits/tmux panes.
+" You can also use <M-h/j/k/l> to resize both tmux panes and vim splits as if
+" they weren the same thing
 
 if exists("g:loaded_tmux_navigator") || &cp || v:version < 700
   finish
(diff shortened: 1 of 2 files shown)


christoomey:
Hi @martin-louazel-engineering, first and foremost, thank you for the kind words and for sharing this work. I'm so glad you've found the plugin valuable 🙂

As for the changes presented here, unfortunately I'm inclined to not include these in this repo. To share some of my thinking:

1. As a baseline, from a maintenance perspective my primary goal is to keep things as simple and stable as possible (as you've learned, the code to make this whole thing work is surprisingly tricky). As such, my bar for considering any new functionality is quite high as I'm very biased towards stability and being able to maintain things (already an uphill battle to be honestA).
1. While the functionality you're sharing here does look great, I personally use https://github.com/roman/golden-ratio and would actively _not_ want this behavior. We could of course make it optional/configurable, but that brings even more complexity, which again I'm hoping to avoid.

All that said, I think you've build something great here and I'd love to support that effort, even if not including the code in this repo. I'd be more than happy to link out to a repo of yours (perhaps a fork, perhaps an augmentation specific to the resizing).

I'm going to close this now based on the above thoughts, but please do let me know if you end up with something I could link to for folks interested in the resizing logic you're sharing here.

martin-louazel-engineering:
Hi @christoomey, thanks a lot for your feedback. I really appreciate it.
Do you have some tips on how I could make it more stable and more like a finished product, so that can be "link-ready" ?
Have a nice day

christoomey:
Hi @martin-louazel-engineering, to your questions

> how I could make it more stable

This, in my experience, is the very tricky part. It's a moving target as vim and tmux versions change over time, and all the more difficult when incorporating more functionality. The main recommendation I can make is to keep this as the core goal (stability) and consider everything else in light of that goal

> more like a finished product, so that can be "link-ready"

Nothing specific here, but the README for this project can serve as a bit of a guide. In you case I'd suggest just referencing this project and then summarizing the extra features you have added to your version.

martin-louazel-engineering:
Hi @christoomey,

Again, thank you very much for your advises.

FYI: I updated the README and made so that my fork can accept new issues for the resizing feature. If you know people that would be interested in this, don't hesitate to redirect them to me. I'll do my best to maintain it.
https://github.com/martin-louazel-engineering/vim-tmux-navigator

Have a great day

PR #320, opened 2022-08-22
Add option to disable wrapping when switch panes

### LINKED ISSUE #233: Add option to prevent wrap around pane movement

maximkulkin:
* Add g:tmux_navigator_no_wrap option to disable wrap around
* Add instructions on how to configure this feature

Based on patch by @heewa

Fixes #233

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,33 @@ With this enabled you can use `<prefix> C-l` to clear the screen.
 
 Thanks to [Brian Hogan][] for the tip on how to re-map the clear screen binding.
 
+#### Disable Wrapping
+
+By default, if you tru to move past the edge of the screen, tmux/vim will
+"wrap" around to the opposite side. To disable this, you'll need to
+configure both tmux and vim:
+
+For vim, you only need to enable this option:
+```vim
+let  g:tmux_navigator_no_wrap = 1
+```
+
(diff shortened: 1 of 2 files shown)


maximkulkin:
@christoomey I have created a fixed conflict-free patch based on PR by @heewa in case original author gave up on it

PR #321, opened 2022-10-06
Optimise `vi` detection logic and use internal macros and commands as much as possible

### LINKED ISSUE #295: `is_vim` does not detect vim running in a subshell

keks24:
Resolves #300 and should also resolve #295.

This pull-request uses:

* The internal macro`#{version}` to determine the `current tmux version`<sup>1</sup>
* `Else clauses` for `if-shell`<sup>2</sup>
* `display-message -p` to output the version in `bash`<sup>3</sup>
* `printf`, `sort` and `head` to check, if the `current tmux version` is higher, than `3.0`. The last two commands come from the package `coreutils`, which has more portability, than `bc`.
* The `vi` detection logic uses a `pseudo substring match` by substituting `g?(view|n?vim?x?)(diff)?$` from `#{pane_current_command}`.<sup>4</sup>

Also everything is refactored to use internal macros and commands as much as possible to cause less load and invoking less subshells.

I was not sure about the `vim`-part, but left a comment there with a snippet, which could be used.

I adapted everything by logic and tested only certain parts of the plugin.

-Ramon

<sup>1</sup>
See `CHANGES FROM 2.3 TO 2.4, 20 April 2017`
https://github.com/tmux/tmux/blob/master/CHANGES

<sup>2</sup>
See `CHANGES FROM 1.5 TO 1.6, 23 January 2012`
https://github.com/tmux/tmux/blob/master/CHANGES

<sup>3</sup>
See `CHANGES FROM 1.1 TO 1.2, 10 March 2010`
https://github.com/tmux/tmux/blob/master/CHANGES

<sup>4</sup>
See `CHANGES FROM 2.1 TO 2.2, 10 April 2016`
https://github.com/tmux/tmux/blob/master/CHANGES

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Add the following to your `~/.tmux.conf` file:
 # Smart pane switching with awareness of Vim splits.
 # See: https://github.com/christoomey/vim-tmux-navigator
 
-# pseudo substring match by substituting the current command from the current commmand string
+# pseudo substring match by substituting "g?(view|n?vim?x?)(diff)?$" from the current commmand string
 bind-key -n C-k if-shell "[ '#{pane_current_command}' != '#{s/g?(view|n?vim?x?)(diff)?$//:#{pane_current_command}}' ]" "send-keys C-k" "select-pane -U"
 bind-key -n C-j if-shell "[ '#{pane_current_command}' != '#{s/g?(view|n?vim?x?)(diff)?$//:#{pane_current_command}}' ]" "send-keys C-j" "select-pane -D"
 bind-key -n C-h if-shell "[ '#{pane_current_command}' != '#{s/g?(view|n?vim?x?)(diff)?$//:#{pane_current_command}}' ]" "send-keys C-h" "select-pane -L"
--- a/README.md
+++ b/README.md
@@ -65,17 +65,21 @@ Add the following to your `~/.tmux.conf` file:
 ``` tmux
 # Smart pane switching with awareness of Vim splits.
 # See: https://github.com/christoomey/vim-tmux-navigator
-is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
-    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|n?vim?x?)(diff)?$'"
-bind-key -n 'C-h' if-shell "$is_vim" 'send-keys C-h'  'select-pane -L'
-bind-key -n 'C-j' if-shell "$is_vim" 'send-keys C-j'  'select-pane -D'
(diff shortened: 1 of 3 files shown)


christoomey:
Hi @keks24, I certainly appreciate the attempts to clean things up, but unfortunately I believe this would be a regression around the handling of subprocesses (e.g. a git alias / subcommand that spawns vim). The use of `ps` was introduced a while back to handle the subprocess consideration and I'd not want to break that functionality.

More generally, while I definitely appreciate the attempts to clean things up and make things more efficient, I'm unlikely to merge a sweeping change like this to the core Vim detection logic. My primary goal these days is to keep things stable and a change like this would require a ton of testing (in addition to the more pointed concerns outlined above).

With all that in mind, I'm going to close this now.

keks24:
I see!

At least it is documented here and others might still have a use of it. :)

-Keks

christoomey:
Absolutely! And the nice thing is that this part is easily swappable by folks so they can grab it if they specifically looking for the efficiency gains 🙂

PR #323, opened 2022-10-15
Map the vim bindings in more modes

druckdev:
Use `(nore)map` to map the functions in visual, select and operator-pending mode instead of only in normal mode.

Use `<C-U>` ([:h c_CTRL-U](https://vimhelp.org/cmdline.txt.html#c_CTRL-U)) to delete the automatically inserted range when pressing `:` in visual mode, as the functions do not support a range.

Update all references in the README and help page.


christoomey:
Thanks for the update, @druckdev!

PR #326, opened 2022-11-08
Prevent control key pass-through when docker is running

anemator:
Prior to this patch we assumed vim would always run on the host system, but docker containers have their own instances of /dev/pts/*, so if vim in say /dev/pts/N was running in tmux which was in a docker container, then we thought vim was running in /dev/pts/N on the host system which caused control keys calls to pass-through as send-keys rather than select-pane on the host system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,27 @@ detail.
 
 [tmate]: http://tmate.io/
 
+### Switching between host panes doesn't work when docker is running
+
+Replace the `is_vim` variable in your `~/.tmux.conf` file with:
+```tmux
+if-shell '[ -f /.dockerenv ]' \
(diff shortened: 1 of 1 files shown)


christoomey (changes requested):
Hi @anemator, I appreciate the update here, but unfortunately I'm inclined to not accept as is. The vim detection logic is super complicated and frankly more brittle than I'd like so I'm super cautious about adding additional logic to it.

I would be open to an update to the troubleshooting section describing the issue and suggesting this as a fix, but I don't want to take this on in the core vim detection code unfortunately.

anemator:
No problem. Just updated my branch. Let me know if you need any more changes.

christoomey:
Thanks, @anemator. Merged now.

PR #363, opened 2023-09-14
added 'v' and 'y' keys for selection and copy in copy-mode-vi

davidfresco:
added binds in copy mode for v to start selection like visual, and y to yank to tmux clipboard


christoomey:
Hey @davidfresco - thanks for sending this in, but looking at it, it seems like these are strictly tmux bindings and not related to the whole vim<>tmux thing. Given that, I'm going to close this (I strive to keep things as simple and focused as possible here).

PR #364, opened 2023-09-16
Don't change active window when fzf is up

### LINKED ISSUE #346: Suggestion: Make this plugin compatible with fzf and similar.
I noticed the `Ctrl+J`/`Ctrl+k` keybindings stopped working on fzf after I had this plugin installed. I added `fzf` to the [is_vim](https://github.com/christoomey/vim-tmux-navigator/blob/master/vim-tmux-navigator.tmux#L5.L6) regex and it fixed the issue.

So I'm wondering: Would it be a good idea to add such exceptions to this plugin?

anakimluke:
FZF uses Control+hjkl key combos for movement and it interferes with this plugin. This commit adds the FZF process name(fzf) to the list of programs that make this plugin not automatically trigger a window change.

This plugin also adds instructions to the readme file in case a user wants to add their own programs to the list, or remove existing ones.

closes #346

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Add the following to your `~/.tmux.conf` file:
 # Smart pane switching with awareness of Vim splits.
 # See: https://github.com/christoomey/vim-tmux-navigator
 is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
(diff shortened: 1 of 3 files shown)


christoomey:
Thanks for the (thorough) update!

PR #371, opened 2023-12-20
Add lazy.nvim instructions

joshmedeski:
The following details are the most simple and efficient way to get this plugin working with the lazy.nvim plugin manager.


christoomey:
Thanks for the update here!

PR #380, opened 2024-03-30
TPM Plugin configure key mapping

Ovascos:
Hello :wave:

as someone who loves TPM but uses a different key binding for pane navigation, I was a bit annoyed that the provided TPM plugin does not support configuring the key mapping for navigation. Of course, someone could resort to copying the provided setting to their .tmux.conf (or even fork this project), but I believe in using plugins to keep the config clean.

Thus, I've updated the plugin to support parameters for configuring the key mapping. It is possible to have multiple keys for each navigation action, as well as disable an individual mapping if desired. Have a look in the updated README to see how it works.

I've tested my changes in tmux version 3.2a, 3.4, as well as 2.9a (as the configuration parsing had some breaking changes in version 3.0). I hope I didn't miss any weird corner case.

### DIFF
--- a/vim-tmux-navigator.tmux
+++ b/vim-tmux-navigator.tmux
@@ -17,6 +17,15 @@ get_tmux_option() {
   fi
 }
 
+bind_key_vim() {
+  local key tmux_cmd is_vim
+  key="$1"
+  tmux_cmd="$2"
+  is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
+      | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|l?n?vim?x?|fzf)(diff)?$'"
+  tmux bind-key -n "$key" if-shell "$is_vim" "send-keys $key" "$tmux_cmd"
+  tmux bind-key -T copy-mode-vi "$key" "$tmux_cmd"
+}
 
 version_pat='s/^tmux[^0-9]*([.0-9]+).*/\1/p'
 move_left="$(get_tmux_option "@vim_navigator_mapping_left" "C-h")"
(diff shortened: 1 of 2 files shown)


christoomey:
Apologies for the delayed reply here, but this looks really great! Thank you for the update -- merging in now.

vim-tmux-navigator.tmux
```diff
@@ -1,25 +1,46 @@
...
+move_prev="$(get_tmux_option "@vim_navigator_mapping_prev" 'C-\')"
+
+for k in $(echo "$move_left");  do bind_key_vim "$k" "select-pane -L"; done
+for k in $(echo "$move_down");  do bind_key_vim "$k" "select-pane -D"; done
+for k in $(echo "$move_up");    do bind_key_vim "$k" "select-pane -U"; done
+for k in $(echo "$move_right"); do bind_key_vim "$k" "select-pane -R"; done
+for k in $(echo "$move_prev");  do bind_key_vim "$k" "select-pane -l"; done
+
+# Restoring clear screen
+tmux bind C-l send-keys 'C-l'
```
jfujitani, line 45: @christoomey  or @Ovascos   
I have prefix C-l bound to resize pane right in .tmux.conf like so:
```
bind -r -T prefix C-h resize-pane -L 3
bind -r -T prefix C-k resize-pane -U 3
bind -r -T prefix C-j resize-pane -D 3
bind -r -T prefix C-l resize-pane -R 3
```
This addition seems to have broken than.  Is there a way I can override this through .tmux.conf?   
Would it be as simple as unbinding this before I set my own binding?

jfujitani: I just moved my key-bind change to after loading this plugin and it worked.  I'm good.  Thanks for the great plugin.

Issue #381, opened 2024-03-31
Cannot switch from nvim to tmux pane


sklbz:
I have neovim with nvchad config.
I added the following lines into `~/.config/nvim/lua/plugins/init.lua`
```Lua
`{
  "christoomey/vim-tmux-navigator",
  cmd = {
    "TmuxNavigateLeft",
    "TmuxNavigateDown",
    "TmuxNavigateUp",
    "TmuxNavigateRight",
    "TmuxNavigatePrevious",
  },
  keys = {
    { "<C-h>", "<cmd><C-U>TmuxNavigateLeft<cr>" },
    { "<C-j>", "<cmd><C-U>TmuxNavigateDown<cr>" },
    { "<C-k>", "<cmd><C-U>TmuxNavigateUp<cr>" },
    { "<C-l>", "<cmd><C-U>TmuxNavigateRight<cr>" },
    { "<C-\\>", "<cmd><C-U>TmuxNavigatePrevious<cr>" },
  },
},
```

The package has been installed but the only command working is `<leader>h`.
The tmux package works just fine.

Here are the informations provided by `neofetch` about my OS.
```bash
OS: Arch Linux on Windows 10 x86_64
Kernel: 5.15.133.1-microsoft-standard-WSL2
Uptime: 10 hours, 43 mins
Packages: 212 (pacman)
Shell: bash 5.2.21
Terminal: Windows Terminal
CPU: Intel i7-9750H (12) @ 2.592GHz
GPU: e35e:00:00.0 Microsoft Corporation Basic Render Driver
Memory: 495MiB / 7865MiB
```

sklbz:
Finally solved the problem by adding some config in init.lua.
I m currently on my phone so can t show the code but will post later

christoomey:
@sklbz appreciate the note and would welcome additional detail when/if you get a moment!

sklbz:
Here are the lines I added to my ~/.config/nvim/init.lua

```lua
vim.keymap.set('n', '<c-k>', ':wincmd k<CR>')
vim.keymap.set('n', '<c-j>', ':wincmd j<CR>')
vim.keymap.set('n', '<c-h>', ':wincmd h<CR>')
vim.keymap.set('n', '<c-l>', ':wincmd l<CR>')
```

ahmedsayedabdelsalam:
still not working, am using nvchad latest version.

Issue #294, opened 2021-05-13
Does not work anymore with tmux 3.2


tklepzig:
It seems with tmux 3.2 jumping out of vim to another tmux pane does not work anymore. Navigating from tmux to vim still works, but not the other way around, simply nothing happens.

tmux: 3.2
vim: 8.2

mhoonjeon:
Same here.

christoomey:
Any errors, or does it just fail to navigate? Also, can you check on the questions in https://github.com/christoomey/vim-tmux-navigator#vim---tmux-doesnt-work?

mhoonjeon:
The `:TmuxNavigatorProcessList` gives me

```
Ss   -zsh
S+   vim
```

Luckily, I ran `Mislav's original external script` as described here at https://github.com/christoomey/vim-tmux-navigator#it-still-doesnt-work, it started working again. Thanks for the advice @christoomey 

@tklepzig You should try this too.

tklepzig:
The `:TmuxNavigatorProcessList` gives me

	Ss   -zsh
	S+   vim
	S+   /Users/me/.asdf/installs/tmux/3.2/bin/tmux

I already hacked a bit with the tmux bindings and when the `$is_vim` check is called via an external script, everything works, maybe that's also the reason why `Mislav's original external script` works. I'm using another tmux related plugin, [vimux](https://github.com/preservim/vimux), which also does not work anymore. It seems there are some breaking changes in version 3.2 of tmux, but I can't find a good candidate in their [changelog](https://raw.githubusercontent.com/tmux/tmux/3.2/CHANGES).

christoomey:
Interesting that Mislav's script works, but the plugin doesn't. That original script used tmux's `pane_current_command`, but this plugin moved off that a few years back to attempt to handle a few edge cases. I'll try to do some more research into this soon, but just wanted to note that distinction in case it points anyone reading this in the right direction.

ds82:
Does not work with `nvim` (0.5.0), but with `vim` (8.2) on the same machine.

adam-gaia:
I'm  in the same boat where navigating from tmux to tmux and navigating from tmux to vim does work, however navigating from vim to tmux does not work.
Using Mislav's script does work for me.

tmux: 3.1c
nvim:  v0.5.0

`:TmuxNavigatorProcessList` outputs
```
S zsh
S nvim
```

jvanbaarsen:
Same problem for me:

[...]

PR #382, opened 2024-04-03
Fix lazy.nvim where most recent stable release doesn't read keybinds (Vim -> tmux broken)

samgarmany:
I had to change my plugin configuration for lazy.nvim for my keybinds within neovim to work, I added event type "VeryLazy" because that is what [this](https://www.lazyvim.org/configuration/general#keymaps) part of the lazyvim documentation seemed to suggest. 

That being said I understand that other users have had success using event type "BufPreRead" and I'm not sure if that one could be better in some way.

**Here are the issues that this relates to:**
#381 
#379 
#294

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ If you are using [lazy.nvim](https://github.com/folke/lazy.nvim). Add the follow
 ```lua
 {
   "christoomey/vim-tmux-navigator",
+  event = "Very Lazy",
   cmd = {
     "TmuxNavigateLeft",
     "TmuxNavigateDown",
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ If you are using [lazy.nvim](https://github.com/folke/lazy.nvim). Add the follow
 ```lua
 {
   "christoomey/vim-tmux-navigator",
-  event = "Very Lazy",
+  event = "VeryLazy",
   cmd = {
     "TmuxNavigateLeft",
     "TmuxNavigateDown",
--- a/README.md
+++ b/README.md
@@ -60,7 +60,8 @@ If you are using [lazy.nvim](https://github.com/folke/lazy.nvim). Add the follow
 ```lua
 {
   "christoomey/vim-tmux-navigator",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -60,6 +60,7 @@ If you are using [lazy.nvim](https://github.com/folke/lazy.nvim). Add the follow
 ```lua
 {
   "christoomey/vim-tmux-navigator",
+  event = "Very Lazy",
```
am3o, line 63:
```suggestion
  event = "VeryLazy",
```
The event has to be called 'VeryLazy' and it just seems like a careless mistake to me.

Notes: I just tried this change and it fixed the local problem. Thanks.

samgarmany: Yikes you’re right, I even got it right in the description. 

I still wonder if bufpreread isn’t better in some way as that’s what I’ve seen more often. VeryLazy just appears in the lazyvim docs.

christoomey:
I've yet to take the jump to full neovim and Lazy.nvim (soon, I hope), but I appreciate the team effort here and references to the other issues so I'm going to optimistically merge this. Thanks for the update!

samgarmany:
I took one more look at the lazy.nvim docs to see if there’s any reason to use the `bufpreread` event type. While I still don’t really know if that would be faster,  I found [this](https://github.com/folke/lazy.nvim?tab=readme-ov-file#lazy-loading) which says that plug-ins should be lazy loaded automatically when they use the `key` key. 

So I now think that the original plugin implementation was fine but there exists a bug in lazy.nvim. I’ll put in an issue report with the lazy team this weekend (if nobody beats me to it) but I think the three best options now are leaving a comment for the user after that line, commenting out the whole line and letting the user uncomment as needed, or not adding the line and waiting for lazy.nvim to fix it.

samgarmany:
I found that the original config works just fine after I changed from NvChad to LazyVim. I'm closing this, I don't know why this happens but it's certainly not necessary for everyone. If someone else finds that keybinds aren't working, try adding that event, but for everyone else it shouldn't be needed.

PR #385, opened 2024-04-15
lazy.nvim instructions for custom keymaps

Tbsnhld:
Instruction on how you can customize the keymappings in lazy.nvim. I had some issues with the old keymappings still being active after changing the keymaps in the lua config. 

Updated the README to include instructions on how to create custom keybindings with lazy.nvim and lua.


christoomey:
I think this is already covered by https://github.com/christoomey/vim-tmux-navigator?tab=readme-ov-file#lazynvim. Could you update this to merge your version into that?

Tbsnhld:
Updated the existing lazy.nvim instructions to include the config part, which ensures that only the specified keybindings are set

PR #390, opened 2024-05-18
use nnoremap instead of noremap

### LINKED ISSUE #388: Conflict with <C-j> and <C-k> keybindings in insert and select modes
I have keymaps for <C-j> and <C-k> to jump to the next and previous snippet parameters in insert and select modes. However, I discovered a conflict with these keybindings when I checked your plugin's keybindings. I realized that these keybindings work in all modes by default. What is the reason for this, and wouldn't it be better if they only worked in normal mode to avoid such conflicts?

MohammadLashkari:
resolve #388, use `nnoremap` instead of `noremap` to avoid conflicts in other modes

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,16 +140,16 @@ Add the following to your `~/.vimrc` to define your custom maps:
 ``` vim
 let g:tmux_navigator_no_mappings = 1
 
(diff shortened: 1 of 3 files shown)


christoomey:
Just a note that I updated to keep the other `noremap` references in sync. Thanks for the note and PR!

PR #392, opened 2024-05-22
document netrw fix

### LINKED ISSUE #379: <C-h> and <C-l> broken inside netrw(neovim). is there any way to fix?

heiskane:
Already answered in #379 but wanted to document it in the readme because i think this is a common use case.
Closes #379

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,20 @@ let g:tmux_navigator_preserve_zoom = 1
 Naturally, if `g:tmux_navigator_disable_when_zoomed` is enabled, this option
 will have no effect.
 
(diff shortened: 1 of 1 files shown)


nfischer:
I wrote up #393 as a way for this plugin to automatically work around the issue. If automatic work around is too risky though, then I think this documentation PR is a really good idea!

christoomey:
@heiskane thanks for writing this up! I ended up opting for the version @nfischer suggested which configures the fix automatically, but wanted to not my appreciation for this writeup nonetheless! Closing this now as #393 has been merged.

PR #394, opened 2024-06-06
Changed all `select-pane` calls to include the -Z option

igor-semyonov:
Hi,

I believe the -Z option for `select-pane` is best as the default. This option makes it so that if you are zoomed into a pane, then you are still zoomed after switching to the new pane.

If you agree, here is a commit making the change.

Best,
Igor


komar007:
Hi @igor-semyonov 

The `-Z` option is already added just below your changes if `g:tmux_navigator_preserve_zoom=1`, so your change breaks `g:tmux_navigator_preserve_zoom=0` setting.

On the vim side of things, you should be able to achieve the effect you're after by just using the flag.

What could be useful is a similar setting for the TPM plugin though, because as far as I can see, it only supports switching panes without `-Z`.

For me, using `-Z` is indeed the default, but then I don't use the TPM plugin...

We would need the maintainer to share some thoughts, @christoomey :)

PR #395, opened 2024-06-11
fix: when in a "pipenv shell" can't move in between different vim/nvim windows

noapoleon:
Simple regex change that adds pipenv to matches
Allows switching panes and vim windows even in a pipenv shell which didn't work before

### DIFF
--- a/vim-tmux-navigator.tmux
+++ b/vim-tmux-navigator.tmux
@@ -3,7 +3,7 @@
 version_pat='s/^tmux[^0-9]*([.0-9]+).*/\1/p'
 
 is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
-    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|l?n?vim?x?|fzf)(diff)?$'"
+    | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|l?n?vim?x?|fzf|pipenv)(diff)?$'"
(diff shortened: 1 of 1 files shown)


christoomey:
Hi @noapoleon, thanks for sharing, but I'm concerned about how broad change would be. Would you be up for switching this from changing the core regex to instead adding a troubleshooting section w/ a summary and note about the altered pattern match?

noapoleon:
> Hi @noapoleon, thanks for sharing, but I'm concerned about how broad change would be. Would you be up for switching this from changing the core regex to instead adding a troubleshooting section w/ a summary and note about the altered pattern match?

Wow I did not expect such a quick reply, sorry for the wait.
Yeah that's totally fine too and I did think about it, my concern was whenever an update is pushed to this repo and the plugin is updated by a plugin manager like Lazy.vim (what I use) it would probably overwrite the changes no?
For now, a troubleshooting section would be fine but what do you think about being able to pass options in a config file like it's done in neovim with a lazy.nvim? Maybe in the form of a list object and whatever is in it could be concatenated with the regex at runtime, that way it doesn't interfere with the local repos. Another approach could be with an environment variable.
I'm not sure how that could be done across different plugin managers honestly and/or tmux, I'm very new to all of this but I thought I'd drop my idea

noapoleon:
Welp.... Just noticed a bug actually.
My change fixed navigation in between different nvim split windows (windows inside a single neovim instance)
BUT if you're using tmux and have say a split pane, on the left pipenv shell, on the right a normal shell, you won't be able to navigate in between them with the keybinds... :[

I'll look into this a bit but no promises

PR #402, opened 2024-08-29
Add documentation to override navigation prefix

thatnerdjosh:
Creating this PR to allow people to override the navigation prefix in tmux.

This can be integrated into the plugin itself along with option to override the key binding. Happy to do this now, or wait and see what ends up happening with https://github.com/christoomey/vim-tmux-navigator/pull/380, as that can change implementation details.

Let me know what you'd recommend.

Thanks :pray: 

Edit: note that this change was inspired by `CTRL+h` being interpreted as the ASCII for backspace, which broke navigation to the left. Making this configurable solves this, and also avoids needing a separate workaround for `CTRL+l` functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Add the following to your `~/.tmux.conf` file:
 version_pat='s/^tmux[^0-9]*([.0-9]+).*/\1/p'
 is_vim="ps -o state= -o comm= -t '#{pane_tty}' \
     | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?g?(view|l?n?vim?x?|fzf)(diff)?$'"
-nav_prefix='M'
+nav_prefix='C'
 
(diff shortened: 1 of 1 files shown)


christoomey:
Hi @thatnerdjosh, I appreciate the note here but I'm inclined to pass on this change. I've just merged the linked TPM change as I think it's useful to expose that functionality since otherwise the keys can't be changed when using TPM install, and the complexity of the change is hidden from the end user.

In this case, users are free to change the keys as they're copying the snippet directly into their tmux conf. I'm inclined to keep the snippet as simple as possible for this install path given all of this, so I'm going to close this one, but again, I appreciate the thought here!

thatnerdjosh:
@christoomey - yup - makes a lot of sense. I made this because I wasn't sure what the direction was for the other PR.

The other one should solve the issue.

However, there was also some minor cleanup in here - I'll make another PR with those changes.

PR #409, opened 2024-10-30
TPM: Fix disabling key bindings

### LINKED ISSUE #408: TPM: Removing key bindings does not work as documented
The README states that to remove a key binding when using TPM, you would have to set the corresponding option to an empty string:

> ```
> set -g @vim_navigator_mapping_prev ""  # removes the C-\ binding
> ```

However this does not work. Instead the default binding (as per this plugin) is used. From https://github.com/christoomey/vim-tmux-navigator/blob/a9b52e7d36114d40350099f254b5f299a35df978/vim-tmux-navigator.tmux#L9-L17 I learned that setting it to `null` would work, and it does.

x3rAx:
The documentation states that for the TPM plugin, key bindings could be disabled by setting the corresponding option to an empty string.

This did not work. Instead `null` could be passed to disable a binding.

This PR changes it so that the function falls back to the default value when the option is not set and allows for using an empty string to disable a mapping. Also, `null` is kept for backward compatibility.

Fixes #408

### DIFF
--- a/vim-tmux-navigator.tmux
+++ b/vim-tmux-navigator.tmux
@@ -4,17 +4,15 @@ get_tmux_option() {
   local option value default
   option="$1"
   default="$2"
-  value=$(tmux show-option -gqv "$option")
+  value="$(tmux show-option -gv "$option" 2>/dev/null || echo "$default")"
 
-  if [ -n "$value" ]; then
-    if [ "$value" = "null" ]; then
+  # Deprecated, for backward compatibility
+  if [[ $value == 'null' ]]; then
       echo ""
-    else
-      echo "$value"
-    fi
-  else
-    echo "$default"
+      return
   fi
+
+  echo "$value"
 }
 
 bind_key_vim() {


christoomey:
Thanks for the updates, @x3rAx!

ghosert:
this fix looks like total break everything for tmux3.2a, it may work for tmux 3.4+ but tmux 3.2a is still default one for Ubuntu 22.04, which a lot of users are using, can you test and fix on tmux3.2a

ghosert:
ctrl + h, j, k, l  totally messed up.

x3rAx:
@ghosert can confirm. As a quick fix, add the following to your tmux config:

```
set -g @vim_navigator_mapping_left "C-h"                                                                                                                                                                         
set -g @vim_navigator_mapping_right "C-l"                                                                                                                                                                        
set -g @vim_navigator_mapping_up "C-k"                                                                                                                                                                           
set -g @vim_navigator_mapping_down "C-j"                                                                                                                                                                         
set -g @vim_navigator_mapping_prev "C-\\"                                                                                                                                                                        
set -g @vim_navigator_prefix_mapping_clear_screen "C-l"
```

x3rAx:
@ghosert Please update to the nevest version, it should now be fixed 😺

ghosert:
Thanks for the quick fix, it's now working. @x3rAx

PR #427, opened 2025-01-24
Close command

joshmedeski:
Hey @christoomey! Thanks for an awesome plugin.

I added a `TmuxNavigateClose` command for:
1. Closing a vim buffer when you're focused on vim
2. Closing a tmux pane when you're focused on a tmux pane

I've found this very helpful for having one binding tied to "closing".

Let me know what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@ configuration described in [this gist][]. When combined with a set of tmux
 key bindings, the plugin will allow you to navigate seamlessly between
 vim and tmux splits using a consistent set of hotkeys.
 
+> [!NOTE]
+> The new `TmuxNavigateClose` command dynamically closes a vim buffer, or a tmux pane, depending on if your focus.
+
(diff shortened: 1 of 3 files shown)


christoomey:
Hey @joshmedeski, thanks for sharing, but unfortunately I think I'll have to pass on including this in the core plugin.

This plugin is surprisingly complicated / somewhat brittle and so I focus mostly on keeping things stable and largely consider it feature complete. Likewise, I'm not keen to claim another global key combo (`C-q` is less used, but not something I'm comfortable taking over).

That said, I really appreciate the work here, and hopefully anyone looking for something similar will be able to reference this PR if they want to implement it themselves.

joshmedeski:
Thanks @christoomey, I totally understand this decision.

I'll be adding more features like this on my fork, thanks for building a helpful plugin 💪

Issue #430, opened 2025-02-15
Enhance terminal mode navigation in vim-tmux-navigator


francescoVaglienti:
## Problem
When using terminal mode (:term) in Vim with vim-tmux-navigator, the Ctrl+h/j/k/l navigation keys don't work properly. This breaks the seamless navigation experience between Vim splits, terminal splits, and tmux panes.

## Solution
Add terminal mode mappings that properly handle navigation while preserving terminal state:

```vim
tnoremap <C-h> <C-w>:<C-U>TmuxNavigateLeft<cr>
tnoremap <C-j> <C-w>:<C-U>TmuxNavigateDown<cr>
tnoremap <C-k> <C-w>:<C-U>TmuxNavigateUp<cr>
tnoremap <C-l> <C-w>:<C-U>TmuxNavigateRight<cr>
```

This retains the expected functionality 

# Benefits

Maintains consistent navigation behavior across all modes
Preserves terminal state when navigating between panes
Works properly with nested Vim instances that contain terminal splits
Seamlessly integrates with existing tmux navigation

# Testing Done

Verified navigation works between:

Vim splits → terminal splits
Terminal splits → Vim splits
Terminal splits → tmux panes
Nested Vim instances with terminal splits

Confirmed terminal state is preserved after navigation

# Note on pull 172
https://github.com/christoomey/vim-tmux-navigator/pull/172 is now almost a decade old and this issue still comes up for default users. Rather than having people figure out how to *create* this functionality, it is much better to have it and just have it turned off by default.

christoomey:
Closed via #431

PR #172, opened 2017-04-10
Add Nvim terminal emulator key mappings

fszymanski:

PR #431, opened 2025-02-15
Add terminal mode navigation support with FZF compatibility fix for issue #430

francescoVaglienti:
The current tmux-navigator plugin doesn't handle terminal mode (term) navigation seamlessly. When using terminal splits in Vim alongside tmux panes, the navigation breaks down. This functionality had been left out because it collided with Vim FZF

Solution
Added terminal mode support with FZF-awareness:

New terminal mode mappings that integrate with existing tmux navigation
Added IsFZF() function to detect when we're in FZF
Used <expr> mappings to conditionally pass through controls to FZF when appropriate

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ mapping.
 
 #### Ignoring programs that use Ctrl+hjkl movement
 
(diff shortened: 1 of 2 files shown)


francescoVaglienti:
I also checked up on https://github.com/christoomey/vim-tmux-navigator/pull/172 to make sure that nothing would actually break. 

When vim-FZF is detected vim-tmux-navigator behavious is stopped and keys are instead passed straight to vim-FZF, you need to then exit vim-FZF to go back to business as usual.

christoomey:
Thanks for the detailed summary and update here, @francescoVaglienti. 👍

PR #433, opened 2025-02-25
fix: ensure `IsFzf` usable when you don't use tmux

phanen:
Related https://github.com/ibhagwan/fzf-lua/issues/1853.
When you don't use tmux, `IsFzf` is undef.


christoomey:
Mind fixing up the conflict here now that I've merged #432

nenahp:
resolved

christoomey:
Thanks for the quick fix here!

PR #443, opened 2025-05-09
fix: adjust testing vim regex

itakashiraiya:


christoomey:
Not sure what this is going for, but I'm quite conservative with changes to the core process checking logic so I'm going to close this.

itakashiraiya:
Sry for not explaining anything, i just changed it bc there was a test that was failing even though it shouldn't. The best solution I found was to make the regex the same as in the core logic, from what I understood this file was just for testing and not any core logic. When I made it the same as the core logic it passed all and only the tests it is supposed to tho.
Feel free to close it again tho, have a good day!

christoomey:
Hi @itakashiraiya, thanks for the clarification and the fix here. I think this same issue you're pointing as is being addressed in a series of PRs by @nfischer (see https://github.com/christoomey/vim-tmux-navigator/commit/64803fdd19a36211eba5e39f86c4cbceae8e1d06 and similar) so I'm going to opt to keep this closed, but again, thanks for the help here

PR #474, opened 2026-05-04
Fix is_vim check to work when docker is running

alebahn:
When docker is running, matching off of TTY can conflict with processes running in docker. This matches based on sessionID.

Full disclosure: the details of this fix came from ChatGPT. It was reviewed and fully tested by me.


christoomey:
Thanks for sharing! I appreciate the note, but I tend to be quite conservative with changes to the core process detection logic and this is enough of an edge case that I'd rather not complicate the default case.

I'd happily merge an addition to the troubleshooting section of the README that describes this as an alternative here.

